PR #131, opened 2019-07-02
fix: 192 bit algorithm suite support in browsers

seebees:
Resolves: # 72

Browsers do not support 192 bit keys.
The proper place for this is in the WebCryptoAlgorithmSuite class.
Move the logic from the CMM to WebCryptoAlgorithmSuite
Add tests to cover the precondition.


modules/material-management/src/web_crypto_algorithms.ts
```diff
@@ -137,6 +141,28 @@ const webCryptoAlgorithms: WebCryptoAlgorithms = Object.freeze({
   [AlgorithmSuiteIdentifier.ALG_AES256_GCM_IV12_TAG16_HKDF_SHA384_ECDSA_P384]: Object.freeze(webCryptoAlgAes256GcmIv12Tag16HkdfSha384EcdsaP384)
 })
 
+/* Web browsers do not support 192 bit key lengths at this time.
+ * To maintain type compatibility and TypeScript happiness between Algorithm Suites
+ * I need to have the same list of AlgorithmSuiteIdentifier.
```
karlw00t: I'm not sure what needs to be kept in sync by this comment. Who is "I"?

seebees: "I" is the author.  You can always check git blame :)

The comment was moved.  Try this:

```
/* Web browsers do not support 192 bit key lengths at this time.
 * To maintain type compatibility and TypeScript happiness between Algorithm Suites
 * the class needs to accept the same list of AlgorithmSuiteIdentifier.
 * The list of Algorithm Suites is maintained here
 * to make sure that the error message is helpful.
 */
```

PR #143, opened 2019-07-10
fix: web-crypto-backend testing

seebees:
More complete tests with stubs

### DIFF
--- a/modules/web-crypto-backend/test/fixtures.ts
+++ b/modules/web-crypto-backend/test/fixtures.ts
@@ -98,3 +98,145 @@ export const subtleFallbackZeroByteEncryptFail = {
 } as any
 
 export const subtleFallbackNoWebCrypto = {} as any
+
+export const fakeWindowIE11OnComplete = {
+  msCrypto: {
+    getRandomValues: (values: Uint8Array) => {
+      return values.fill(1)
+    },
+    subtle: {
+      decrypt () {
+        const obj = {} as any
+        setTimeout(() => {
+          obj.result = true
+          obj.oncomplete()
+        })
+        return obj
+      },
+      digest () {
+        const obj = {} as any
+        setTimeout(() => {
+          obj.result = true
+          obj.oncomplete()
+        })
+        return obj
+      },
+      encrypt () {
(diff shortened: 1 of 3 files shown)


modules/web-crypto-backend/test/fixtures.ts
```diff
@@ -98,3 +98,145 @@ export const subtleFallbackZeroByteEncryptFail = {
...
+} as any
+
+export const fakeWindowIE11OnError = {
+  msCrypto: {
+    getRandomValues: (values: Uint8Array) => {
+      return values.fill(1)
+    },
+    subtle: {
+      decrypt () {
+        const obj = {} as any
```
caitlin-tibbetts: I don't know much about JavaScript/TypeScript, but I'm noticing that a lot of these function definitions are exactly the same, could you make one function and call it whenever needed?

modules/web-crypto-backend/test/synchronous_random_values.test.ts
```diff
@@ -18,11 +18,28 @@
 import { expect } from 'chai'
 import 'mocha'
 import { synchronousRandomValues } from '../src/index'
+import sinon from 'sinon'
+import * as browserWindow from '@aws-sdk/util-locate-window'
```
caitlin-tibbetts: Be careful about doing a * import statement, could involve importing unneeded things

seebees: :) . However, I need to stub the module.  That is why I'm doing it

modules/web-crypto-backend/test/fixtures.ts
seebees: I could.  An this is effectively what I'm doing in the code that uses this.  I often chose to be more verbose in tests so that the test code and the code that is tested are not equally abstract...

modules/web-crypto-backend/test/synchronous_random_values.test.ts
caitlin-tibbetts: Oh okay, so it's specific for the sinon import (per a Google search), that makes more sense

PR #148, opened 2019-07-15
fix: Encryption Context changes

### LINKED ISSUE #54: Encryption context should be required everywhere but the initial user entry points
There are a lot of points in the codebase that accept an optional encryption context and default to an empty object if so. We should only do that default value assignment once, at the top-level user interface. Everywhere else, we should require that it is actually provided.

seebees:
resolves #54

1. browser/node encrypt function public api should match
1. Encryption context should _only_ be optional in the public interface
1. Encryption context moved into cryptographic materials

Internally, it is much more consistent to require an encryption context.
Moving the encryption context onto the cryptographic materials
brings them in line with the python keyring implementation.
It also makes control of the encryption context in the CMM
and **not** the keyrings very clear in the interface.

NOTE: encrypt in encrypt-node now expects `encryptionContext` instead of `context`.
This is **ONLY** done because the project is still in beta.


modules/cache-material/test/clone_cryptographic_material.test.ts
```diff
@@ -46,7 +46,7 @@ const cryptoKey: any = { type: 'secret', algorithm: { name: webCryptoSuite.encry
 
 describe('cloneMaterial', () => {
   it('clone NodeEncryptionMaterial', () => {
-    const material = new NodeEncryptionMaterial(nodeSuite)
+    const material = new NodeEncryptionMaterial(nodeSuite, { some: 'context' })
```
karlw00t: What error is thrown when no context is provided? Should we test for that?

seebees: You can not create Encryption Materials without encryption context.

https://github.com/awslabs/aws-encryption-sdk-javascript/blob/master/modules/material-management/test/cryptographic_material.test.ts#L430

The public api (encrypt/decrypt) should be the _only_ place that encryption context is optional.

mattsb42-aws:
> NOTE: encrypt in encrypt-node now expects encryptionContext instead of context.
This is ONLY done because the project is still in beta.

Agreed, and I think this was the right call.

Just calling out here, though, that even though we're in beta we will still need to update the version appropriately for a breaking change (so this will be `0.2.0`)

modules/material-management-browser/src/browser_cryptographic_materials_manager.ts
```diff
@@ -90,22 +89,25 @@ export class WebCryptoDefaultCryptographicMaterialsManager implements WebCryptoM
...
+    const { signatureCurve: namedCurve } = suite
 
-    /* Precondition: The algorithm suite specification must support a signatureCurve to extract a verification key. */
-    if (!namedCurve) return material
+    /* Check for early return (Postcondition): The WebCryptoAlgorithmSuite specification must support a signatureCurve to extract a verification key. */
+    if (!namedCurve) return new WebCryptoDecryptionMaterial(suite, encryptionContext)
 
     /* Precondition: WebCryptoDefaultCryptographicMaterialsManager If the algorithm suite specification requires a signatureCurve a context must exist. */
-    if (!context) throw new Error('Context does not contain required public key.')
+    if (!encryptionContext) throw new Error('Context does not contain required public key.')
```
mattsb42-aws:
```suggestion
    if (!encryptionContext) throw new Error('Encryption context does not contain required public key.')
```

modules/material-management-browser/test/browser_cryptographic_materials_manager.test.ts
```diff
@@ -203,11 +203,11 @@ describe('WebCryptoDefaultCryptographicMaterialsManager', () => {
       some: 'context'
     }
 
-    const test = await cmm.getEncryptionMaterials({ suite, encryptionContext })
-    expect(Object.keys(test.context)).lengthOf(2)
-    if (!test.material.signatureKey) throw new Error('I should never see this error')
-    expect(test.context).to.have.haveOwnProperty(ENCODED_SIGNER_KEY).and.to.equal(toBase64(test.material.signatureKey.compressPoint))
-    expect(test.context).to.have.haveOwnProperty('some').and.to.equal('context')
+    const { material } = await cmm.getEncryptionMaterials({ suite, encryptionContext })
```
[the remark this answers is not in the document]
mattsb42-aws: > I know that makes it a bit difficult if we wanted to add something else in to this response later

For the record, this is exactly what the encryption materials structure is for. ;) That's what we need to modify if we want to return more things.

modules/material-management-node/src/node_cryptographic_materials_manager.ts
```diff
@@ -48,73 +48,75 @@ export class NodeDefaultCryptographicMaterialsManager implements NodeMaterialsMa
...
+  _initializeDecryptionMaterial (suite: NodeAlgorithmSuite, encryptionContext: EncryptionContext) {
+    const { signatureCurve: namedCurve } = suite
 
     /* Check for early return (Postcondition): The algorithm suite specification must support a signatureCurve to load a signature key. */
-    if (!namedCurve) return material
+    if (!namedCurve) return new NodeDecryptionMaterial(suite, encryptionContext)
 
     /* Precondition: NodeDefaultCryptographicMaterialsManager If the algorithm suite specification requires a signatureCurve a context must exist. */
-    if (!context) throw new Error('Context does not contain required public key.')
+    if (!encryptionContext) throw new Error('Context does not contain required public key.')
```
mattsb42-aws:
```suggestion
    if (!encryptionContext) throw new Error('Encryption context does not contain required public key.')
```

modules/material-management/src/cryptographic_material.ts
```diff
@@ -22,6 +22,23 @@ import { NodeAlgorithmSuite } from './node_algorithms'
...
+ * SDK and you are not developing your own CMMs and/or keyrings, you do not
+ * need to use it and you should not do so.
+ *
+ * The CryptographicMaterial's purpose is to bind together all the required elements for
+ * encrypting or decrypting a payload.
+ * The functional data key (unencrypted or CryptoKey) is the most sensitive data and needs to
+ * be protected.  The longer this data persists in memory the
+ * greater the opportunity to be invalidated.  Because
+ * a Caching CMM exists is it important to insure that the
+ * unencrypted data key and it's meta data can not be manipulated,
```
mattsb42-aws:
```suggestion
 * unencrypted data key and its meta data can not be manipulated,
```

seebees:
@mattsb42-aws 
> Just calling out here, though, that even though we're in beta we will still need to update the version appropriately for a breaking change (so this will be `0.2.0`)

To be clear, I'm using: https://www.conventionalcommits.org/en/v1.0.0-beta.4/ so if it was called out as breaking in the commit message, the MAJOR version would be bumped automatically

modules/kms-keyring/test/kms_keyring.ondecrypt.test.ts
```diff
@@ -46,7 +46,7 @@ describe('KmsKeyring: _onDecrypt',
       const clientProvider: any = () => {
         return { decrypt }
         function decrypt ({ CiphertextBlob, EncryptionContext, GrantTokens }: any) {
-          expect(EncryptionContext === context).to.equal(true)
+          expect(EncryptionContext).to.deep.equal(context)
```
ragona: Nitpick: Since you just ran around doing s/context/encryptionContext/ everywhere else, you might consider doing that here and in the other tests where it's still `context`. I don't think I normally would have even noticed but after reviewing that change in 26 files it stuck out. :P

modules/material-management/test/keyring.test.ts
```diff
@@ -46,7 +45,7 @@ describe('Keyring', () => {
   it('onEncrypt calls _onEncrypt', async () => {
     const suite = new NodeAlgorithmSuite(AlgorithmSuiteIdentifier.ALG_AES128_GCM_IV12_TAG16)
     const { keyLengthBytes } = suite
-    const m = new NodeEncryptionMaterial(suite)
+    const m = new NodeEncryptionMaterial(suite, {})
```
ragona: Nitpick: Would prefer a more descriptive name here.

modules/material-management/test/multi_keyring.test.ts
```diff
@@ -215,10 +215,10 @@ describe('MultiKeyring: onDecrypt', () => {
     const suite = new NodeAlgorithmSuite(AlgorithmSuiteIdentifier.ALG_AES128_GCM_IV12_TAG16)
     const unencryptedDataKey = new Uint8Array(suite.keyLengthBytes)
     const [edk0, keyringTrace0] = makeEDKandTrace(0)
-    const material = new NodeDecryptionMaterial(suite)
+    const material = new NodeDecryptionMaterial(suite, {})
 
     const generator = keyRingFactory({
-      async onDecrypt (material: NodeDecryptionMaterial /*, encryptedDataKeys: EncryptedDataKey[], context?: EncryptionContext */) {
+      async onDecrypt (material: NodeDecryptionMaterial /*, encryptedDataKeys: EncryptedDataKey[] */) {
```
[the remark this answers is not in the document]
seebees: For me, this is similar to named parameters.  Ideally the _both_ the function declaration _and_ the call site should be descriptive of intention.

In this case, the function is "cut off" because I'm only using _some_ of the arguments.  My Typescript config is relatively strict and wants every defined variable to be used.  So I have to remove the arguments because they are not being used.  However when copying this code around or trying to understand what/how it is used, it is useful to see the full definition.

modules/material-management/test/keyring.test.ts
seebees: What are you looking for?  Defining the variable as opposed to just passing an object lateral?

modules/kms-keyring/test/kms_keyring.ondecrypt.test.ts
seebees: This is valid.  The examples are another case of this.... Sigh.  I agree, but I think I'll come back to this later.  Since this is large enough it will be easier to change words after a stable position.

#160

ragona: +1, definitely not worth holding this up.

modules/material-management/test/keyring.test.ts
ragona: Oh, sorry, I meant `m` not `{}`. Not a big deal.

modules/material-management/test/multi_keyring.test.ts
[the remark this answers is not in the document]
ragona: Makes sense!

modules/material-management/src/cryptographic_material.ts
```diff
@@ -22,6 +22,23 @@ import { NodeAlgorithmSuite } from './node_algorithms'
...
+ * developers of CMMs and keyrings only. If you are a user of the AWS Encryption
+ * SDK and you are not developing your own CMMs and/or keyrings, you do not
+ * need to use it and you should not do so.
+ *
+ * The CryptographicMaterial's purpose is to bind together all the required elements for
+ * encrypting or decrypting a payload.
+ * The functional data key (unencrypted or CryptoKey) is the most sensitive data and needs to
+ * be protected.  The longer this data persists in memory the
+ * greater the opportunity to be invalidated.  Because
+ * a Caching CMM exists is it important to insure that the
```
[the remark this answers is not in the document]
seebees: fixed

PR #149, opened 2019-07-16
fix: frame length can not be 0

### LINKED ISSUE #129: FrameLength can not be zero, and must be less than Maximum. FRAME_SIZE
Both of these need to be preconditions in `serialize` at a minimum

seebees:
resolves #129

The other ESDK’s use frameLength === 0 to indicate non-framed content.
The JS ESDK does not support encrypting non-framed content.
A frameLength of 0 will crash the process.
Add condition and test.

### DIFF
--- a/modules/encrypt-browser/src/encrypt.ts
+++ b/modules/encrypt-browser/src/encrypt.ts
@@ -63,7 +63,7 @@ export async function encrypt (
   { suiteId, encryptionContext, frameLength = FRAME_LENGTH }: EncryptInput = {}
 ): Promise<EncryptResult> {
   /* Precondition: The frameLength must be less than the maximum frame size for browser encryption. */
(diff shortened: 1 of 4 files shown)


modules/encrypt-browser/src/encrypt.ts
```diff
@@ -60,6 +62,10 @@ export async function encrypt (
   plaintext: Uint8Array,
   { suiteId, encryptionContext, frameLength = FRAME_LENGTH }: EncryptInput = {}
 ): Promise<EncryptResult> {
+
+  /* Precondition: The frameLength must be less than the maximum frame size for browser encryption. */
+  needs(frameLength > 0 && Maximum.FRAME_SIZE >= frameLength, 'frameLength out of bounds')
```
mattsb42-aws: Can we add those bounds to the error message?

seebees: fixed, in both... :)

PR #153, opened 2019-07-18
feat: Encryption tests for integration-node

seebees:
https://github.com/awslabs/aws-crypto-tools-test-vector-framework defines an encryption manifest.
This manifest should be used to maintain cross compatibility of language implementations.

Then integrate this into the CI integration tests.

### DIFF
--- a/modules/integration-node/src/integration_tests.ts
+++ b/modules/integration-node/src/integration_tests.ts
@@ -15,13 +15,28 @@
 
 import {
   TestVectorInfo, // eslint-disable-line no-unused-vars
-  getTestVectorIterator
-} from './get_test_iterator'
-import { decryptMaterialsManagerNode } from './decrypt_materials_manager_node'
-import { decrypt } from '@aws-crypto/client-node'
+  getDecryptTestVectorIterator
+} from './get_decrypt_test_iterator'
+import {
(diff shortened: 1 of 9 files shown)


modules/integration-node/src/integration_tests.ts
```diff
@@ -34,30 +49,80 @@ export async function testVector ({ name, keysInfo, plainTextStream, cipherStrea
...
+  let failureCount = 0
+  for (const test of tests) {
+    if (testName) {
+      if (test.name !== testName) continue
+    }
+    const { result, name, err } = await testDecryptVector(test)
+    if (result) {
+      console.log({ name, result })
+    } else {
+      if (err && notSupportedDecryptMessages.includes(err.message)) {
```
karlw00t: Do we want to enumerate the vectors that are not supported? Right now, if something returns not supported, but should be, it would slip by. That being said, this can go into a new issue, and doesn't need to be fixed here.

karlw00t: or is that what "tolerateFailures" does?

seebees: tolerateFailures is just a count of failures to tolerate.  CI should be 0.

Your comment is true.  The hope is to come up with a results manifest.  That could be compared and determine success.

PR #159, opened 2019-07-23
feat: Encryption tests for integration-browser

seebees:
https://github.com/awslabs/aws-crypto-tools-test-vector-framework defines an encryption manifest.
This manifest should be used to maintain cross compatibility of language implementations.

Then integrate this into the CI integration tests.
Also update package.json to better organize parameters.

### DIFF
--- a/modules/integration-browser/src/build_encrypt_fixtures.ts
+++ b/modules/integration-browser/src/build_encrypt_fixtures.ts
@@ -44,7 +44,9 @@ export async function buildEncryptFixtures (fixtures: string, manifestFile: stri
   Object
     .keys(plaintexts)
     .forEach(name => {
-      /* https://github.com/awslabs/aws-crypto-tools-test-vector-framework/blob/master/features/0003-awses-message-encryption.md#plaintexts */
+      /* Generate random bites as per spec.
+       * See: https://github.com/awslabs/aws-crypto-tools-test-vector-framework/blob/master/features/0003-awses-message-encryption.md#plaintexts
+       */
       plaintextBytes[name] = randomBytes(10).toString('base64')
     })
 
--- a/modules/integration-browser/src/build_encrypt_fixtures.ts
+++ b/modules/integration-browser/src/build_encrypt_fixtures.ts
@@ -1,0 +1,121 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use
+ * this file except in compliance with the License. A copy of the License is
+ * located at
+ *
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {
+  EncryptManifestList, // eslint-disable-line no-unused-vars
+  KeyList, // eslint-disable-line no-unused-vars
+  KeyInfoTuple // eslint-disable-line no-unused-vars
+} from './types'
+import { randomBytes } from 'crypto'
+import {
+  AlgorithmSuiteIdentifier, // eslint-disable-line no-unused-vars
+  EncryptionContext // eslint-disable-line no-unused-vars
+} from '@aws-crypto/client-browser'
+import { URL } from 'url'
+import { readFileSync, writeFileSync } from 'fs'
+import got from 'got'
+
+export async function buildEncryptFixtures (fixtures: string, manifestFile: string, keyFile: string, testName: string, slice: string) {
+  const [start = 0, end = 9999] = (slice || '').split(':').map(n => parseInt(n, 10))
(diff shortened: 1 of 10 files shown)


modules/integration-browser/src/cli.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  writeFileSync(`${fixtures}/decrypt_tests.json`, JSON.stringify([]))
+  writeFileSync(`${fixtures}/encrypt_tests.json`, JSON.stringify([]))
+  writeFileSync(`${fixtures}/decrypt_oracle.json`, JSON.stringify(decryptOracle))
+
+  if (command === 'decrypt') {
+    const { vectorFile } = argv
+    // @ts-ignore
+    await buildDecryptFixtures(fixtures, vectorFile, testName, slice)
+  } else if (command === 'encrypt') {
```
johnwalker: Why ignore the await?

Similarly, I noticed a few eslint disables through the tests, for unused variables. Can we keep these linters and checks on?

modules/integration-browser/src/integration.encrypt.test.ts
```diff
@@ -0,0 +1,62 @@
...
+ *
+ *     http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* eslint-env jasmine */
```
johnwalker: Is there any way to apply this to a directory?

modules/integration-browser/src/types.ts
```diff
@@ -13,11 +13,21 @@
  * limitations under the License.
  */
 
-export interface ManifestList {
-  manifest: Manifest
+import { EncryptionContext } from '@aws-crypto/client-browser' // eslint-disable-line no-unused-vars
```
johnwalker: It looks like it's being used. Would you remove the eslint disable?

modules/integration-browser/src/integration.encrypt.test.ts
```diff
@@ -0,0 +1,62 @@
...
+            'Content-Type': 'application/octet-stream',
+            'Accept': 'application/octet-stream'
+          },
+          body: cipherMessage
+        })
+        const body = await response.arrayBuffer()
+        needs(response.ok, `Failed to decrypt: ${toUtf8(body)}`)
+        expect(plainText).toEqual(new Uint8Array(body))
+      } catch (e) {
+        if (!notSupportedMessages.includes(e.message)) throw e
```
ragona: I don't suppose there's a better way to handle specific error types that isn't just string parsing? It looks like if one of those error messages changes the logic here breaks.

modules/integration-browser/src/build_encrypt_fixtures.ts
```diff
@@ -0,0 +1,121 @@
...
+import { randomBytes } from 'crypto'
+import {
+  AlgorithmSuiteIdentifier, // eslint-disable-line no-unused-vars
+  EncryptionContext // eslint-disable-line no-unused-vars
+} from '@aws-crypto/client-browser'
+import { URL } from 'url'
+import { readFileSync, writeFileSync } from 'fs'
+import got from 'got'
+
+export async function buildEncryptFixtures (fixtures: string, manifestFile: string, keyFile: string, testName: string, slice: string) {
```
johnwalker: Would you add some comments to this? I need help understanding the choice of start, end, 10, the randomBytes, and what the goal of the method is.

modules/integration-browser/src/build_encrypt_fixtures.ts
```diff
@@ -0,0 +1,121 @@
...
+  const [start = 0, end = 9999] = (slice || '').split(':').map(n => parseInt(n, 10))
```
ragona: Nitpick: This smells funny.

modules/integration-browser/src/integration.encrypt.test.ts
seebees: Not that I know of

modules/integration-browser/src/types.ts
seebees: It is technically not being used.  From the JavaScript side, this is not used.  There should be a better solution to this eventually, but not at this time.

modules/integration-browser/src/integration.encrypt.test.ts
seebees: There maybe.  But that would be an extensive change to the ESDK that I may make...

modules/integration-browser/src/build_encrypt_fixtures.ts
seebees: Very funny.  Just me working around some cmd line parsing :(
Humans are difficult interfaces...

modules/integration-browser/src/cli.ts
seebees: It is not being used from a JS perspective, only Typescript type...

The function does not return anything, so there is nothing to "get"

modules/integration-browser/src/build_encrypt_fixtures.ts
seebees: added

PR #163, opened 2019-07-23
test: Testing cacheing CMM

seebees:
Simple refactor to better test the node caching CMM

# Check any applicable:

### DIFF
--- a/modules/caching-materials-manager-node/test/sha512.test.ts
+++ b/modules/caching-materials-manager-node/test/sha512.test.ts
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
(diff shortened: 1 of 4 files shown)


modules/caching-materials-manager-node/test/sha512.test.ts
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+
+/* eslint-env mocha */
+
+import { expect } from 'chai'
+import 'mocha'
+import { sha512 } from '../src/sha512'
+
+// sha512('asdf')
```
karlw00t: Should this be base64(sha512('asdf'))?

seebees: You are not wrong.  I left the base64 off because the important part was the 'asdf'

PR #165, opened 2019-07-26
fix: Multi keyrings should not require a generator

seebees:
If a multi keyring did not have a generator,
but was passed material with an unencrypted data key,
it would throw.
This is incorrect.
Better code, tests, and error messages.

# Check any applicable:

### DIFF
--- a/modules/material-management/test/multi_keyring.test.ts
+++ b/modules/material-management/test/multi_keyring.test.ts
@@ -202,7 +202,7 @@ describe('MultiKeyring: onEncrypt', () => {
     })
 
     const mkeyring = new MultiKeyringNode({  children:[ child ] })
-    const material = new NodeEncryptionMaterial(suite) //.setUnencryptedDataKey(unencryptedDataKey, keyringTrace0)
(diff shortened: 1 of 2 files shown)


modules/material-management/test/multi_keyring.test.ts
```diff
@@ -189,7 +189,21 @@ describe('MultiKeyring: onEncrypt', () => {
...
+    const suite = new NodeAlgorithmSuite(AlgorithmSuiteIdentifier.ALG_AES128_GCM_IV12_TAG16)
+    const child = keyRingFactory({
+      async onEncrypt (material: NodeEncryptionMaterial) {
+        return material
+      },
+      onDecrypt: never
+    })
+
+    const mkeyring = new MultiKeyringNode({  children:[ child ] })
+    const material = new NodeEncryptionMaterial(suite) //.setUnencryptedDataKey(unencryptedDataKey, keyringTrace0)
```
karlw00t: Why are we keeping this commented code around?

seebees: removed

PR #171, opened 2019-07-30
test: Condition tests for web-crypto-backend

seebees:

# Check any applicable:


modules/web-crypto-backend/test/backend-factory.test.ts
```diff
@@ -189,3 +195,25 @@ describe('webCryptoBackendFactory', () => {
...
 })
+
+describe('getNonZeroByteBackend', () => {
+  it('gets a subtle backend', () => {
+    const test = getNonZeroByteBackend(fixtures.subtleFallbackSupportsZeroByteGCM)
+    expect(test === fixtures.subtleFallbackSupportsZeroByteGCM.subtle).to.equal(true)
+  })
+
+  it('Precondition: A backend must be passed to get a non zero byte backend.', () => {
+    expect(() => getNonZeroByteBackend(false)).to.throw('No supported backend.')
```
lizroth: Is there a convenient way to make the meaning of the `false` more intuitive to the reader (e.g. by naming it, or adding a comment for what it does)?

modules/web-crypto-backend/test/backend-factory.test.ts
```diff
@@ -189,3 +195,25 @@ describe('webCryptoBackendFactory', () => {
...
+})
+
+describe('getZeroByteSubtle', () => {
+  it('gets a subtle backend', () => {
+    const test = getZeroByteSubtle(fixtures.subtleFallbackSupportsZeroByteGCM)
+    expect(test === fixtures.subtleFallbackSupportsZeroByteGCM.subtle).to.equal(true)
+  })
+
+  it('Precondition: A backend must be passed to get a zero byte backend.', () => {
+    expect(() => getZeroByteSubtle(false)).to.throw('No supported backend.')
```
lizroth: Ditto

modules/web-crypto-backend/test/backend-factory.test.ts
seebees: Hum... These functions operate on backends returned.  I can add a comment.  But the idea is that if a backend does not exist, I return false to be explicit.  These helper functions are just handling that...

lizroth: We chatted about this offline and I can't think of a non-confusing way to express what I want here, so I would say let's just leave it.

PR #173, opened 2019-07-30
test: Condition tests for serialize

seebees:

# Check any applicable:


modules/serialize/src/aad_factory.ts
```diff
@@ -44,7 +44,7 @@ export function aadFactory (fromUtf8: (input: string) => Uint8Array) {
           ? ContentAADString.FINAL_FRAME_STRING_ID
           : ContentAADString.FRAME_STRING_ID
       default:
-        throw new Error('bad')
+        throw new Error('Unrecognized context type')
```
mattsb42-aws:
```suggestion
        throw new Error('Unrecognized content type')
```

mattsb42-aws: Also, should this be a particular type of error?

modules/serialize/src/decode_body_header.ts
```diff
@@ -148,7 +148,8 @@ export function decodeFinalFrameBodyHeader (buffer: Uint8Array, headerInfo: Head
 
   /* The precondition SEQUENCE_NUMBER_END: Uint32(FFFF) is handled above. */
   const sequenceEnd = dataView.getUint32(readPos, false) // big endian
-  needs(sequenceEnd === SequenceIdentifier.SEQUENCE_NUMBER_END, '')
+  /* Postcondition: sequenceEnd must be SEQUENCE_NUMBER_END. */
+  needs(sequenceEnd === SequenceIdentifier.SEQUENCE_NUMBER_END, 'Malformed sequenceEnd')
```
mattsb42-aws: This seems too specific and unlikely that an arbitrary developer is going to understand what it means.
```suggestion
  needs(sequenceEnd === SequenceIdentifier.SEQUENCE_NUMBER_END, 'Malformed final frame: Invalid sequence number end value')
```

modules/serialize/src/aad_factory.ts
seebees: I have not gotten to "types of errors" _yet_.  Error codes are a thing :(

PR #175, opened 2019-07-30
test: Condition tests for decrypt-node

seebees:

# Check any applicable:
- [x] Were any files moved? Moving files changes their URL, which breaks all hyperlinks to the files.


modules/decrypt-node/test/decipher_stream.test.ts
```diff
@@ -0,0 +1,132 @@
...
+      sequenceNumber: 1,
+      isFinalFrame: false
+    }
+    test._onBodyHeader(bodyInfo)
+
+    await expect(test._onAuthTag(Buffer.from([]), () => {})).to.rejectedWith(Error, 'AuthTag before frame.')
+  })
+
+  it('Precondition: I must have received all content for this frame.', async () => {
+    /* The fact that I can not figure out how to test this,
```
lizroth: Let's cut an issue tracking figuring this out and link it here?

seebees: It is not really testable.  I'm noting this in code and handling this differently.
Basically there is a closure that holds this state.  And I do not see _how_, short of magically changing raw RAM, to make this fail in order to test it...

PR #177, opened 2019-07-30
test: Condition tests for caching-materials-manager-browser

seebees:

# Check any applicable:


modules/caching-materials-manager-browser/test/caching_materials_manager_browser.test.ts
```diff
@@ -59,4 +59,28 @@ describe('WebCryptoCachingMaterialsManager', () => {
     expect(test._maxBytesEncrypted).to.equal(maxBytesEncrypted)
     expect(test._maxMessagesEncrypted).to.equal(maxMessagesEncrypted)
   })
+
+  it('Precondition: A partition value must exist for WebCryptoCachingMaterialsManager.', () => {
```
lavaleri: Does this test also need to be added to caching_materials_manager_node.test.ts ?

modules/caching-materials-manager-browser/test/caching_materials_manager_browser.test.ts
```diff
@@ -59,4 +59,28 @@ describe('WebCryptoCachingMaterialsManager', () => {
...
+    const maxAge = 10
+    const test = new WebCryptoCachingMaterialsManager({
+      backingMaterials: keyring,
+      cache,
+      maxAge
+    })
+    /* Binary data is being transformed to utf-8.
+     * This means it is highly likely to be less than 64 characters long.
+     */
+    expect(test._partition).to.be.a('string').and.to.have.length.greaterThan(50)
```
lavaleri: What is the likelihood that 64 bytes of random data -> UTF-8 will be less than 50 characters? Enough that we will "never" see this test fail?

modules/caching-materials-manager-browser/test/caching_materials_manager_browser.test.ts
seebees: Yes, and it is in a different PR

modules/caching-materials-manager-browser/test/caching_materials_manager_browser.test.ts
seebees: Sigh... I "fixed" this in https://github.com/awslabs/aws-encryption-sdk-javascript/pull/178/files#diff-2e3dd129c30c2285d72b61362efd41b9R58

Once that is merged.  I'll update the test... utf8 on random data is.... sub optimal.

modules/caching-materials-manager-browser/test/caching_materials_manager_browser.test.ts
```diff
@@ -59,4 +59,28 @@ describe('WebCryptoCachingMaterialsManager', () => {
...
+    const keyring = new TestKeyring()
+    const cache = 'cache' as any
+    const maxAge = 10
+    const test = new WebCryptoCachingMaterialsManager({
+      backingMaterials: keyring,
+      cache,
+      maxAge
+    })
+    /* Binary data is being transformed to utf-8.
+     * This means it is highly likely to be less than 64 characters long.
```
lavaleri: This comment is no longer needed

PR #181, opened 2019-08-03
fix: maxBodySize can not short circuit on frameLengh

seebees:
In the case of a framed message that encrypts less than the frameLenght,
maxBodySize was doing the wrong thing.

Remove the check.
Add tests vectors that cover the case.

# Check any applicable:


modules/decrypt-node/test/decrypt.test.ts
```diff
@@ -83,4 +83,21 @@ describe('decrypt', () => {
       { encoding: 'base64' }
     )).to.rejectedWith(Error, 'Invalid Signature')
   })
+
```
lavaleri: No unit tests for ParseHeaderStream?

modules/decrypt-node/src/parse_header_stream.ts
```diff
@@ -35,20 +34,12 @@ interface HeaderState {
   buffer: Buffer
 }
 
-export interface ParseHeaderOptions {
-  maxBodySize?: number
-}
-
 export class ParseHeaderStream extends PortableTransformWithType {
```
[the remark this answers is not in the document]
seebees: I'm not sure I'm following...

https://github.com/awslabs/aws-encryption-sdk-javascript/pull/181/files#diff-a2a13589c6d6b1804302d56578057d38R43

modules/decrypt-node/test/decrypt.test.ts
seebees: Specifically, no.  But implicitly, the `stream` test in this file will exercise just about everything.

modules/decrypt-node/src/parse_header_stream.ts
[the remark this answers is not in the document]
lavaleri: Oh weird, for some reason I missed that change. Ignore me.

modules/decrypt-node/test/decrypt.test.ts
lavaleri: Unless there is some reason why unit tests for ParseHeaderStream specifically would not be useful, we should add them.
Either in this PR, or (because the logic added in this PR seems to be sufficiently tested below) tracked in a new issue.

seebees: #188

PR #185, opened 2019-08-04
fix: Conditions for materials-management

seebees:
Every condition needs a test.
This also uncovered some missing conditions and tests.

* If a multi keyring has errors AND fails to decrypt the EDK it should throw and error.
* Lower the test_conditions threshold

# Check any applicable:


modules/material-management/src/cryptographic_material.ts
```diff
@@ -58,7 +58,7 @@ function portableTimingSafeEqual (a: Uint8Array, b: Uint8Array) {
    * Side channel attacks are pernicious and subtle.
    */
   eval('') // eslint-disable-line no-eval
-  /* Check for early return (Postcondition): Size is well-know information
+  /* Check for early return (Postcondition) UNTESTED: Size is well-know information.
```
lavaleri: Why is this untested? Is there a reason this is not testable, or is adding the test outside the scope of this PR?

modules/material-management/src/cryptographic_material.ts
```diff
@@ -338,8 +338,9 @@ export function decorateCryptographicMaterial<T extends CryptographicMaterial<T>
     udkForVerification.fill(0)
     unencryptedDataKeyZeroed = true
 
-    /* Postcondition: Both unencryptedDataKey and udkForVerification must be either set or unset.
+    /* Postcondition UNTESTED: Both unencryptedDataKey and udkForVerification must be either set or unset.
```
lavaleri: Why is this untested? Is there a reason this is not testable, or is adding the test outside the scope of this PR?

modules/material-management/src/keyring.ts
```diff
@@ -50,7 +50,7 @@ export abstract class Keyring<S extends SupportedAlgorithmSuites> {
      */
     needs(material === _material, 'New EncryptionMaterial instances can not be created.')
 
-    /* Postcondition: If this keyring generated data key, it must be the right length.
+    /* Postcondition UNTESTED: If this keyring generated data key, it must be the right length.
```
lavaleri: Why is this untested? Is there a reason this is not testable, or is adding the test outside the scope of this PR?

modules/material-management/src/multi_keyring.ts
```diff
@@ -113,19 +113,29 @@ function buildPrivateOnDecrypt<S extends SupportedAlgorithmSuites> () {
...
+        keyringError = true
       }
     }
+
+    /* Postcondition: Need either a valid data key or no errors.
+     * If I have a key, decrypt errors should be ignored.
+     * However, if I was unable to decrypt a key AND I have errors,
+     * these errors should bubble up.
+     */
+    needs(material.hasValidKey() || !keyringError, 'Keyring errors and no keyring was successfully able to decrypt the encrypted data key.')
```
lavaleri: naive JS question: How does JS handle the bubbling up of errors? If this needs condition is not met due to an error caught on L122, how does the error (and possibly others) propagate up, and how is that error handled?

modules/material-management/test/algorithm_suites.test.ts
```diff
@@ -28,10 +28,22 @@ describe('AlgorithmSuiteIdentifier', () => {
 describe('AlgorithmSuite', () => {
   it('should not allow an instance', () => {
     // @ts-ignore Trying to test something that Typescript should deny...
-    expect(() => new AlgorithmSuite()).to.throw()
+    expect(() => new AlgorithmSuite()).to.throw('new AlgorithmSuite is not allowed')
```
lavaleri: Another niave JS question, what is going on here: https://github.com/awslabs/aws-encryption-sdk-javascript/blob/master/modules/material-management/src/algorithm_suites.ts#L105

Is this preventing the AlgorithmSuite constructor from being called? Is this because you're trying to ensure that only extensions of AlgorithmSuite are actually constructed?

modules/material-management/test/cryptographic_material.test.ts
```diff
@@ -97,12 +97,24 @@ describe('decorateCryptographicMaterial', () => {
...
+  })
+
+  it(`Precondition: If the unencryptedDataKey has not been set, it should not be settable later.
+      Precondition: If the udkForVerification has not been set, it should not be settable later.`, () => {
     const suite = new NodeAlgorithmSuite(AlgorithmSuiteIdentifier.ALG_AES128_GCM_IV12_TAG16)
     const test = decorateCryptographicMaterial((<any>{ suite, keyringTrace: [] }), KeyringTraceFlag.WRAPPING_KEY_GENERATED_DATA_KEY)
     test.zeroUnencryptedDataKey()
     const dataKey = new Uint8Array(suite.keyLengthBytes).fill(1)
     const trace = { keyNamespace: 'k', keyName: 'k', flags: KeyringTraceFlag.WRAPPING_KEY_GENERATED_DATA_KEY }
+    // It is very hard to test this perfectly.  However, this tests the spirit.
```
lavaleri: What is the "spirit" that is actually being tested here, and what test is missing for the above preconditions?

This looks to me like were testing the postcondition at https://github.com/awslabs/aws-encryption-sdk-javascript/blob/master/modules/material-management/src/cryptographic_material.ts#L341

I'm not familiar with the precise definition of pre/post condition as used in this codebase, but if we aren't checking something and throwing an error on the precondition does it make sense to call it a precondition?

modules/material-management/test/cryptographic_material.test.ts
```diff
@@ -309,6 +352,27 @@ describe('decorateWebCryptoMaterial', () => {
     expect(() => test.setCryptoKey(key2, trace)).to.throw()
   })
 
+  it('Precondition: If the CryptoKey is the only version, the trace information must be set here.', () => {
```
lavaleri: It is unclear to me what "If the CryptoKey is the only version" means. Looking at the code, this should be a precondition about the keyring trace during setCryptoKey if the udk isn't already set.

package.json
```diff
@@ -26,7 +26,7 @@
     "integration-node-encrypt": "npm run build; lerna run integration_node --stream --no-prefix -- -- encrypt -m $npm_package_config_encryptManifestList -k $npm_package_config_encryptKeyManifest -o $npm_package_config_decryptOracle",
     "integration-node": "run-s integration-node-*",
     "integration": "run-s integration-*",
-    "test_conditions": "./util/bootstrap_tsconfig 23"
+    "test_conditions": "./util/bootstrap_tsconfig 2"
```
lavaleri: What does this option do?

modules/raw-rsa-keyring-browser/test/get_import_options.test.ts
```diff
@@ -27,7 +27,7 @@ chai.use(chaiAsPromised)
 const { expect } = chai
 
 describe('flattenMixedCryptoKey', () => {
-  it('Check for early return (Postcondition): empty inputs should return an empty array.', () => {
+  it('Check for early return (Postcondition): nothing is an empty array.', () => {
```
lavaleri: I think this wording changed recently. Did you intend to change it back?

modules/material-management/src/multi_keyring.ts
```diff
@@ -113,19 +113,29 @@ function buildPrivateOnDecrypt<S extends SupportedAlgorithmSuites> () {
...
       } catch (e) {
-      // there should be some debug here?  or wrap?
       // Failures onDecrypt should not short-circuit the process
       // If the caller does not have access they may have access
       // through another Keyring.
+        keyringError = true
       }
     }
+
+    /* Postcondition: Need either a valid data key or no errors.
```
lavaleri: This logic is in line with what is currently in the spec PR. However, I either want someone familiar with the C multikeyring implementation to either give this logic the OK, or give the current wording in the spec the OK, before I approve this. I will tag someone in the spec CR on this issue.

modules/material-management/src/cryptographic_material.ts
seebees: Mostly because constant time function are not verifiable (as I understand it).  So this function is not exported so it can not be used elsewhere.

The fact that I return in this place is something that needs to be explained, but I don't even know how I would verify this one way or another...

modules/material-management/src/cryptographic_material.ts
seebees: Yes:  It is not clear how this could ever happen, unless someone is manipulating the OS...

The values are private, I have no way to manipulate the values to verify that the function will indeed fail in this case.

modules/material-management/src/keyring.ts
seebees: This is covered by the materials themselves.  The comment is there to make sure that people know it needs to be verified.  But there is no code to be tested...

modules/material-management/src/multi_keyring.ts
seebees: Thanks :)

modules/material-management/src/multi_keyring.ts
seebees: The errors from L122 are lost and not wrapped.  That is not a bad idea to add, but it will require some other re-work

modules/material-management/test/algorithm_suites.test.ts
seebees: > Is this preventing the AlgorithmSuite constructor from being called? 

yes

> Is this because you're trying to ensure that only extensions of AlgorithmSuite are actually constructed?

yes

when you create an object with `new` there is a special property `constructor` that is just what it says.  :)

modules/material-management/test/cryptographic_material.test.ts
seebees: If the unencryptedDataKey has not been set, it should not be settable later.

i.e. if you create materials, then `test.zeroUnencryptedDataKey()`, you should not be able to set an unencrypted data key later.

modules/material-management/test/cryptographic_material.test.ts
seebees: If you only set a CryptoKey.  And do not set an unencrypted data key.  This happens in browsers when an EDK can be "unwrapped".  The API lets you do this unwrapping and put the data key into an un exportable CryptoKey.  This means that the unencrypted data key is never available to the javascript.

modules/raw-rsa-keyring-browser/test/get_import_options.test.ts
seebees: Hum, the test must match the condition.  You may be right....

package.json
seebees: I am trying to get rid of it.  I just lets the test not fail if there are "exceptions" I want to remove it so that no one even sees it and thinks, "Hey, I can get aways with not writing a test by setting this..."  But as I get to that point, I want to have it running...

modules/material-management/test/cryptographic_material.test.ts
lavaleri: gotcha :+1:

modules/material-management/test/cryptographic_material.test.ts
lavaleri: Ok, I see the behavior now. This test depends on setUnencryptedDataKey maintaining a behavior where it checks whether the unencryptedDataKey/udkForVerification has been set. This makes the test more brittle than I'd like for the precondition as a whole, but at the very least we ensure with this test that the case really we care about ("you can't set after zeroing) is taken care of.

PR #186, opened 2019-08-06
fix: caching cmm export and material

seebees:
* WebCryptoDecryptionMaterial do not have an unencrypted data key
This is because the CrypoKey offers better security,
and some unwrapping algorithms can directly return a CryptoKey
without exposing the unencrypted data key.
Update test for WebCryptoDecryptionMaterial.

* Caching CMMs need to be able to create a local cryptographic materials cache
export the function.

# Check any applicable:


modules/cache-material/test/clone_cryptographic_material.test.ts
```diff
@@ -87,13 +87,12 @@ describe('cloneMaterial', () => {
   })
 
   it('clone WebCryptoDecryptionMaterial', () => {
+    /* WebCryptoDecryptionMaterial do not have an unencrypted data key. */
     const material = new WebCryptoDecryptionMaterial(webCryptoSuite, { some: 'context' })
-      .setUnencryptedDataKey(udk128, trace)
```
lavaleri: I know this change fixes the cloning so that it handles the WebCryptoDecryptionMaterials case, but is setting an unencryptedDataKey on WebCryptoDecryptionMaterials something that should be allowed in the first place?

modules/caching-materials-manager-browser/src/index.ts
```diff
@@ -14,3 +14,4 @@
  */
 
 export * from './caching_materials_manager_browser'
+export { getLocalCryptographicMaterialsCache } from '@aws-crypto/cache-material'
```
lavaleri: Where are these exports being used? Should they be included in a PR containing changes which include them?

seebees: You can not create a local cache easily without these.  The should always have been exported...  I found this "bug" in writing the examples.  I can put it somewhere else, but I felt like they were in the same world.

modules/cache-material/test/clone_cryptographic_material.test.ts
seebees: This may be true.  I'd have to think about it.  But it would also be a VERY large change.  Having it _should_ not hurt anything...

lavaleri: If there is a potential for this being a sharp edge, can we either document this somewhere (maybe CryptoKey or WebCryptoDecryptionMaterial) or create an issue for further investigation?

lavaleri: Discussed offline, we do not see this as a sharp edge. Investigation for how to make this easy to maintain is part of https://github.com/awslabs/aws-encryption-sdk-javascript/issues/74

PR #193, opened 2019-08-19
fix: BYTES_PER_KEY value

seebees:
Maximum number of bytes which are allowed to be encrypted under a single data key
The _real_ maximum is 2 ** 63 - 1,
However Javascript can only perform safe operations on values
up to Number.MAX_SAFE_INTEGER === 9007199254740991 === 2 ** 53 - 1.
e.g
Number.MAX_SAFE_INTEGER + 1 === Number.MAX_SAFE_INTEGER + 2 => true
Number.MAX_SAFE_INTEGER + 1 > Number.MAX_SAFE_INTEGER + 2 => false
Number.MAX_SAFE_INTEGER + 1 < Number.MAX_SAFE_INTEGER + 2 => false

This means that after 2 ** 53 - 1 the process of accumulating a byte count
will never yield an accurate comparison and so, never halt.

The choice here to use 2 ** 53 - 1 instead of Number.MAX_SAFE_INTEGER is deliberate.
This is because in the future Number.MAX_SAFE_INTEGER could be raised to 2 ** 66
or some value larger 2 ** 63.

# Check any applicable:

### DIFF
--- a/modules/serialize/src/identifiers.ts
+++ b/modules/serialize/src/identifiers.ts
@@ -78,7 +78,7 @@ export enum Maximum {
    * The choice here to use 2 ** 53 - 1 instead of Number.MAX_SAFE_INTEGER is deliberate.
    * This is because in the future Number.MAX_SAFE_INTEGER could be raised to 2 ** 66
    * or some value larger 2 ** 63.
-   * 
+   *
    * This is *not* the maximum amount of data that can be encrypted under a single data key
    * or under a single AES operation.
    * The maximum amount of data that can be safely encrypted under a single AES operation is 2 ** 36 -32.
--- a/modules/serialize/src/identifiers.ts
+++ b/modules/serialize/src/identifiers.ts
@@ -62,7 +62,7 @@ Object.freeze(SequenceIdentifier)
 export enum Maximum {
   // Maximum number of messages which are allowed to be encrypted under a single cached data key
   MESSAGES_PER_KEY = 2 ** 32, // eslint-disable-line no-unused-vars
-  /* Maximum number of bytes which are allowed to be encrypted
+  /* Maximum number of bytes that are allowed to be encrypted
    * under a single cached data key across messages.
    * The _real_ maximum is 2 ** 63 - 1,
    * However Javascript can only perform safe operations on values
--- a/modules/serialize/src/identifiers.ts
+++ b/modules/serialize/src/identifiers.ts
@@ -62,8 +62,23 @@ Object.freeze(SequenceIdentifier)
 export enum Maximum {
   // Maximum number of messages which are allowed to be encrypted under a single cached data key
   MESSAGES_PER_KEY = 2 ** 32, // eslint-disable-line no-unused-vars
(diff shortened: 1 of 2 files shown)


modules/serialize/src/identifiers.ts
```diff
@@ -62,8 +62,23 @@ Object.freeze(SequenceIdentifier)
 export enum Maximum {
   // Maximum number of messages which are allowed to be encrypted under a single cached data key
   MESSAGES_PER_KEY = 2 ** 32, // eslint-disable-line no-unused-vars
-  // Maximum number of bytes which are allowed to be encrypted under a single cached data key
-  BYTES_PER_KEY = 2 ** 63 - 1, // eslint-disable-line no-unused-vars
+  /* Maximum number of bytes which are allowed to be encrypted under a single data key
```
mattsb42-aws: I want to make sure we're clear in any comments/names/etc that this is the limit applied to cached data key reuse *across messages*, not data key use *within a single message*.

seebees: pushed update

modules/serialize/src/identifiers.ts
```diff
@@ -62,8 +62,34 @@ Object.freeze(SequenceIdentifier)
 export enum Maximum {
   // Maximum number of messages which are allowed to be encrypted under a single cached data key
   MESSAGES_PER_KEY = 2 ** 32, // eslint-disable-line no-unused-vars
-  // Maximum number of bytes which are allowed to be encrypted under a single cached data key
-  BYTES_PER_KEY = 2 ** 63 - 1, // eslint-disable-line no-unused-vars
+  /* Maximum number of bytes which are allowed to be encrypted
+   * under a cached single data key across messages.
+   * The _real_ maximum is 2 ** 63 - 1,
```
mattsb42-aws: I still think this is misleading; we're giving too little information for this to actually be accurate.

My gut is that we should probably not talk a lot about the *why* in a comment in a source file; that is better served by proper documentation in our concepts overviews.

There are three value at play here:

**Cryptographic Safety**

Maximum number of bytes that we will allow within a single AES operation: 2^63 - 32.
We chose this limit based on the mathematical characteristics of the AES algorithm.
This is also the maximum number of bytes that we allow in an unframed message because the entire body is a single AES operation.

**Message Format Bounds**

Maximum number of bytes that we will allow in a single frame: 2^32 - 1.
This limit is defined and enforced by the message format.

Maximum frames that we will allow in a single framed message: 2^32 - 1.
This limit is defined and enforced by the message format.

Maximum number of bytes we will allow in a single framed message: (2^32 - 1) * (2^32 - 1).
This limit is a consequence of the previous two.

**Caching Constraints**

Maximum number of messages we will allow to be protected under a single data key: 2^32.

Maximum number of bytes we will allow to be protected under a single data key before we stop reusing that data key: 2^63 - 1.
* As you pointed out, this is not exactly a limit on the number of bytes protected; it is more correctly a point at which we will cease reuse.

PR #196, opened 2019-08-20
fix: ENCODED_SIGNER_KEY in encryption context

seebees:
A CryptographicMaterialsManager can change entries to the encryptionContext
but changing these values has consequences.
The DefaultCryptographicMaterialsManager uses the value in the encryption context
to store public signing key.
If the caller is using this value in their encryption context the Default CMM
is probably not the CMM they want to use.

# Check any applicable:

### DIFF
--- a/modules/material-management-browser/src/browser_cryptographic_materials_manager.ts
+++ b/modules/material-management-browser/src/browser_cryptographic_materials_manager.ts
@@ -49,11 +49,23 @@ export class WebCryptoDefaultCryptographicMaterialsManager implements WebCryptoM
   async getEncryptionMaterials ({ suite, encryptionContext }: WebCryptoEncryptionRequest): Promise<WebCryptoEncryptionMaterial> {
     suite = suite || new WebCryptoAlgorithmSuite(AlgorithmSuiteIdentifier.ALG_AES256_GCM_IV12_TAG16_HKDF_SHA384_ECDSA_P384)
 
+    /* Precondition: WebCryptoDefaultCryptographicMaterialsManager should reserve the ENCODED_SIGNER_KEY constant from @aws-crypto/serialize.
+     * A CryptographicMaterialsManager can change entries to the encryptionContext
+     * but changing these values has consequences.
+     * The DefaultCryptographicMaterialsManager uses the value in the encryption context to store public signing key.
+     * If the caller is using this value in their encryption context the Default CMM is probably not the CMM they want to use.
+     */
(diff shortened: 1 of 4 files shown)


modules/material-management-browser/src/browser_cryptographic_materials_manager.ts
```diff
@@ -49,11 +49,23 @@ export class WebCryptoDefaultCryptographicMaterialsManager implements WebCryptoM
   async getEncryptionMaterials ({ suite, encryptionContext }: WebCryptoEncryptionRequest): Promise<WebCryptoEncryptionMaterial> {
     suite = suite || new WebCryptoAlgorithmSuite(AlgorithmSuiteIdentifier.ALG_AES256_GCM_IV12_TAG16_HKDF_SHA384_ECDSA_P384)
 
+    /* Precondition: WebCryptoDefaultCryptographicMaterialsManager should reserve the ENCODED_SIGNER_KEY constant from @aws-crypto/serialize.
```
mattsb42-aws:
```suggestion
    /* Precondition: WebCryptoDefaultCryptographicMaterialsManager must reserve the ENCODED_SIGNER_KEY constant from @aws-crypto/serialize.
```

modules/material-management-node/src/node_cryptographic_materials_manager.ts
```diff
@@ -49,11 +49,23 @@ export class NodeDefaultCryptographicMaterialsManager implements NodeMaterialsMa
   async getEncryptionMaterials ({ suite, encryptionContext }: NodeEncryptionRequest): Promise<NodeEncryptionMaterial> {
     suite = suite || new NodeAlgorithmSuite(AlgorithmSuiteIdentifier.ALG_AES256_GCM_IV12_TAG16_HKDF_SHA384_ECDSA_P384)
 
+    /* Precondition: NodeDefaultCryptographicMaterialsManager should reserve the ENCODED_SIGNER_KEY constant from @aws-crypto/serialize.
```
mattsb42-aws:
```suggestion
    /* Precondition: NodeDefaultCryptographicMaterialsManager must reserve the ENCODED_SIGNER_KEY constant from @aws-crypto/serialize.
```

modules/material-management-browser/src/browser_cryptographic_materials_manager.ts
seebees: fixed, and the test is updated...

modules/material-management-node/src/node_cryptographic_materials_manager.ts
seebees: fixed, and the test is updated...

PR #200, opened 2019-08-28
feat: Suport Node.js crypto KeyObjects

### LINKED ISSUE #74: Support KeyObjects in Node.js
From cursory investigation of Node Crypo module native code implementation, Node makes use of smart pointers to zero memory (for symmetric keys) or call OpenSSL helpers (pkey_free()) to clear private key memory contents when keys are no longer needed.

https://nodejs.org/api/crypto.html#crypto_class_keyobject

seebees:
resolves #74

KeyObjects were introduced in Node.js v11.
They wrap the Buffer in an object that has
even better security properties than an isolated buffer.

# Check any applicable:

### DIFF
--- a/modules/material-management-node/src/material_helpers.ts
+++ b/modules/material-management-node/src/material_helpers.ts
@@ -216,7 +216,7 @@ export function nodeKdf (material: NodeEncryptionMaterial|NodeDecryptionMaterial
     info instanceof Uint8Array,
     'Invalid HKDF values.'
   )
-  /* The unwrap is done once we *know* that a KDF in required.
+  /* The unwrap is done once we *know* that a KDF is required.
    * If we unwrapped before everything will work,
    * but we may be creating new copies of the unencrypted data key (export).
    */
--- a/modules/material-management-node/src/material_helpers.ts
+++ b/modules/material-management-node/src/material_helpers.ts
@@ -15,6 +15,9 @@
 
 import {
   needs, NodeEncryptionMaterial, NodeDecryptionMaterial,
+  unwrapDataKey,
+  wrapWithKeyObjectIfSupported,
+  AwsEsdkKeyObject, // eslint-disable-line no-unused-vars
   NodeHash // eslint-disable-line no-unused-vars
 } from '@aws-crypto/material-management'
 import {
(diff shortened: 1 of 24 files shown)


modules/material-management/src/types.ts
```diff
@@ -65,3 +65,27 @@ export type DecryptionMaterial<Suite> =
   Suite extends NodeAlgorithmSuite ? NodeDecryptionMaterial :
   Suite extends WebCryptoAlgorithmSuite ? WebCryptoDecryptionMaterial :
   never
+
+export type AwsEsdkKeyObjectType = 'secret' | 'public' | 'private'
```
lavaleri: What is the purpose of these exports?

modules/material-management/test/cryptographic_material.test.ts
```diff
@@ -163,7 +180,14 @@ describe('decorateCryptographicMaterial', () => {
     material.setUnencryptedDataKey(dataKey, { keyNamespace: 'k', keyName: 'k', flags: KeyringTraceFlag.WRAPPING_KEY_GENERATED_DATA_KEY })
     const test = material.getUnencryptedDataKey()
     test[0] = 12
-    expect(() => material.getUnencryptedDataKey()).to.throw()
```
lavaleri: If I understand the code below, we are testing that one way to try to modify a KeyObject udk doesn't work. The purpose of this test should be to test that if it does get modified, then we fail. If it should be impossible to modify the KeyObject, then perhaps the below is fine for the KeyObject case, however I still think we should check that is the udk is not a KeyObject then it still correctly fails if modified.

seebees: In the non-KeyObject case `getUnencryptedDataKey` will throw before the `dataKey` is even returned.  In this case we are dealing with Uint8Array (raw memory).  The cryptographic material keeps an independent copy of the `dataKey` and compares it to the `dataKey` that it is going to return.

This is because _all_ returned instances are references to the same memory so anyone _could_ have manipulated the memory.

modules/material-management/src/types.ts
seebees: Need to add comments.  Sorry.  This is because there are different versions of Node.js.  If I use/export the v11/v12 version then Typescript will complain if consumers of this package use any other version of Node.js (and we must support older versions.)

modules/material-management-node/src/material_helpers.ts
```diff
@@ -212,10 +216,17 @@ export function nodeKdf (material: NodeEncryptionMaterial|NodeDecryptionMaterial
     info instanceof Uint8Array,
     'Invalid HKDF values.'
   )
+  /* The unwrap is done once we *know* that a KDF in required.
```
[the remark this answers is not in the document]
seebees: updated

PR #201, opened 2019-08-28
feat: Remove unencryptedDataKeyLength

seebees:
It is a feature that is not being used.
This was created to let callers confirm that the unencrypted data key matched the algorithm suite,
but this checking is done directly into the cryptographic material now.
The `hasUnencryptedDataKey` is the best way to check for existence,
and the materials themselves will ensure that algorithm suite specifications are meet.

# Check any applicable:


seebees:
NOTE: This breaking change is **ONLY** done because the project is in beta.
Otherwise there would be a major version bump.

PR #202, opened 2019-08-28
fix: Zero length frames in old version of Node

### LINKED ISSUE #199: Zero length frames in Node.js v 10.9 and older fail
The Decipher stream wants to be initialized with a `update` call before `final` is called.

seebees:
resolves #199

In Node.js versions 10.9 and older will fail to decrypt if decipher.update is not called.
https://github.com/nodejs/node/pull/22538 fixes this.

If the content is empty, push an empty buffer.

# Check any applicable:


mattsb42-aws:
Do we have a test the makes sure this fixes the problem?

seebees:
I manually tested this.  At this time CI does not test this version.  I have also added an issue to expand testing across node.js versions #206

PR #212, opened 2019-09-16
fix: Better error messageing

### LINKED ISSUE #152: Better error messaging when no keyring is found.
Currently if no configured keyring is found to unwrap any of the encrypted data keys,
the CMM will throw an `unencryptedDataKey has not been set` error.
While this error is accurate, it is unhelpful.

seebees:
resolves #152

When attempting to decrypt a set of encrypted data keys,
if any attempt is successful,
then the entire operation should be considered successful.
However if no data key can be obtained,
and there were errors,
these errors should be visible to the caller.

An excellent example is attempting to decrypt
with a KMS CMK alias arn.
The KMS Keyring will be unable to decrypt,
but was returning no error.
This resulted the Error 'Unencrypted data key is invalid.'
This is because the default CMM sees that the material
does not have any unencrypted data key.

A better error message would be the one from KMS in this case.

Updating with tests both the KMS Keyrings,
as well as the MultiKeyring.

# Check any applicable:


modules/kms-keyring/src/kms_keyring.ts
```diff
@@ -203,6 +206,20 @@ export function KmsKeyringClass<S extends SupportedAlgorithmSuites, Client exten
...
+      /* Postcondition: A CMK must provide a valid data key.
+       * If I have a data key,
+       * decrypt errors can be ignored.
+       * However, if I was unable to decrypt a data key AND I have errors,
+       * these errors should bubble up.
+       * Otherwise, the only error customers will see is that
+       * the material does not have an unencrypted data key.
+       * So I return a concatenated Error message
+       */
+      needs(material.hasValidKey() || (!material.hasValidKey() && !cmkErrors.length)
```
lavaleri: I'm a bit unsure if this is the right behavior. Because the spec is silent about error handling, we should go with what makes sense for this implementation while being cognizant of what the C implementation is doing.

In C, OnDecrypt will always return success except possibly on line https://github.com/aws/aws-encryption-sdk-c/blob/master/aws-encryption-sdk-cpp/source/kms_keyring.cpp#L133
However, it will still log an error in these cases:
- if no key was found by the end of the function, these errors:
  - errors from KMS Decrypt if in "non-discovery" mode
  - a parsed invalid EDK ARN error (https://github.com/aws/aws-encryption-sdk-c/blob/master/aws-encryption-sdk-cpp/source/kms_keyring.cpp#L93)

As such, the C implementation has the following behavior: "If in discovery mode, do not log errors from KMS Decrypt". Should the JS implementation also have this behavior? I.e. Should we bubble up CMK errors from KMS discovery keyrings to the user? Or should it be silent about those errors like in the C implementation?

modules/kms-keyring/test/kms_keyring.ondecrypt.test.ts
```diff
@@ -138,10 +138,20 @@ describe('KmsKeyring: _onDecrypt',
...
         return { decrypt }
-        function decrypt () {
-          throw new Error('failed to decrypt')
+        function decrypt ({ CiphertextBlob, EncryptionContext, GrantTokens }: any) {
+          if (edkCount === 0) {
+            edkCount += 1
+            throw new Error('failed to decrypt')
+          }
+          expect(EncryptionContext).to.deep.equal(context)
+          expect(GrantTokens).to.equal(grantTokens)
```
lavaleri: I think we are missing a test for the case when `!material.hasValidKey() && !cmkErrors.length`

modules/kms-keyring/src/kms_keyring.ts
```diff
@@ -203,6 +206,20 @@ export function KmsKeyringClass<S extends SupportedAlgorithmSuites, Client exten
         return material
       }
 
+      /* Postcondition: A CMK must provide a valid data key.
```
lavaleri: Is this the right wording if it is valid for no CMKs to produce a valid data key as long as there are no errors?

modules/material-management/src/multi_keyring.ts
```diff
@@ -113,19 +113,37 @@ function buildPrivateOnDecrypt<S extends SupportedAlgorithmSuites> () {
...
-      // through another Keyring.
+      /* Failures onDecrypt should not short-circuit the process
+       * If the caller does not have access they may have access
+       * through another Keyring.
+       */
+        childKeyringErrors.push(e)
       }
     }
+
+    /* Postcondition: A child keyring must provide a valid data key.
```
lavaleri: Same wording issue here.

modules/material-management/test/multi_keyring.test.ts
```diff
@@ -358,6 +358,23 @@ describe('MultiKeyring: onDecrypt', () => {
     expect(unwrapDataKey(test.getUnencryptedDataKey())).to.deep.equal(unencryptedDataKey)
     expect(called).to.equal(true)
   })
+
```
lavaleri: Same here, I think we want to test the case where `!material.hasValidKey() && !childKeyringErrors.length`

PR #213, opened 2019-09-16
fix: plaintextLength must be enforced

seebees:
If the user has expressed a plaintext length,
this value MUST not be exceed.

# Check any applicable:


modules/encrypt-node/src/framed_encrypt_stream.ts
```diff
@@ -49,11 +50,17 @@ const ioTick = () => new Promise(resolve => setImmediate(resolve))
...
+export function getFramedEncryptStream (getCipher: GetCipher, messageHeader: MessageHeader, dispose: Function, plaintextLength?: number) {
   let accumulatingFrame: AccumulatingFrame = { contentLength: 0, content: [], sequenceNumber: 1 }
   let pathologicalDrain: Function = noop
   const { frameLength } = messageHeader
 
+  /* Precondition: plaintextLength must be withing bounds.
+   * The Maximum.BYTES_PER_CACHED_KEY_LIMIT is set to be within Number.MAX_SAFE_INTEGER
+   * See serialize/identifiers.ts enum Maximum for more details.
+   */
+  needs(!plaintextLength || (plaintextLength >= 0 && Maximum.BYTES_PER_CACHED_KEY_LIMIT >= plaintextLength), 'plaintextLength out of bounds.')
```
mattsb42-aws: Per offline discussion, `Maximum.BYTES_PER_CACHED_KEY_LIMIT` is the incorrect limit to apply here.
It should be `Maximum.FRAME_COUNT * Maximum.FRAME_SIZE`.
Because of the JS number limitations, this ends up needing to have the same numerical value, but we need to define it as a separate limit identifier.

PR #216, opened 2019-09-16
fix: prototype inheritance bug in decodeEncryptionContext

PetarMax:
If the serialised encryption context contains a key-value pair where the key coincides with a property of Object.prototype, decodeEncryptionContext will incorrectly throw 'Error: Duplicate encryption context key value'.

The fix is to create the encryptionContext using Object.create(null) instead of the standard object initialiser.

Nota bene: This fix disables the standard calling of Object.prototype functions on encryption contexts returned by decodeEncryptionContext. In particular, it is no longer possible to call encryptionContext.hasOwnProperty(prop). However, due to the fix, one can use the simpler encryptionContext[prop] instead.

# Check any applicable:


modules/serialize/test/deserialize_factory.test.ts
```diff
@@ -78,6 +78,29 @@ describe('deserializeFactory:decodeEncryptionContext', () => {
     expect(test).to.have.property('information')
       .and.to.eql('\u00bd + \u00bc = \u00be')
   })
+
+  it('Keys may be properties of Object.prototype, decodeEncryptionContext has to succeed', () => {
+    const { decodeEncryptionContext } = deserializeFactory(toUtf8, WebCryptoAlgorithmSuite)
+
+    /* Single key-value pair */
```
seebees: Can you move this to the `fixtures.ts` file?  This is a good test vector to have, and is a little more portable :)

PetarMax: Yes, I have added the new test vector in `fixtures.ts` and adjusted the tests accordingly.

PR #217, opened 2019-09-18
fix: version and type are required by the message format

### LINKED ISSUE #209: Verify the version and type in deserializeMessageHeader
Re #208, the error manifested as an invalid algorithm suite,
but the underlying issue was the message format.
If the version and type were verified as well,
the error message would have been better.

seebees:
resolves #209

The version and type are required values.
This error will help clarify malformed messages.
Especially string messages that have been encoded
and the proper encoding has not been passed.

# Check any applicable:


modules/serialize/test/fixtures.ts
```diff
@@ -29,6 +29,14 @@ export function suiteIdNotValidMessageHeader () {
...
+}
+
+export function typeNotValidMessageHeader () {
+  return new Uint8Array([ 1, 256, 0, 20, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 0, 43, 0, 2, 0, 11, 105, 110, 102, 111, 114, 109, 97, 116, 105, 111, 110, 0, 12, 194, 189, 32, 43, 32, 194, 188, 32, 61, 32, 194, 190, 0, 4, 115, 111, 109, 101, 0, 6, 112, 117, 98, 108, 105, 99, 0, 2, 0, 12, 194, 189, 32, 43, 32, 194, 188, 32, 61, 32, 194, 190, 0, 8, 102, 105, 114, 115, 116, 75, 101, 121, 0, 5, 1, 2, 3, 4, 5, 0, 12, 194, 189, 32, 43, 32, 194, 188, 32, 61, 32, 194, 190, 0, 9, 115, 101, 99, 111, 110, 100, 75, 101, 121, 0, 5, 6, 7, 8, 9, 0, 2, 0, 0, 0, 0, 12, 0, 0, 16, 0 ])
+}
+
```
mattsb42-aws: Can we get a vector and test for the base64-encoded case?

PR #224, opened 2019-10-03
docs: Add versioning file

seebees:

# Check any applicable:

### DIFF
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -11,7 +11,7 @@ Major version changes are significant and expected to break backwards compatibil
 Minor version changes will not break compatibility between the previous minor versions;
 to do so is a bug.
 Encryption SDK changes will also involve addition of optional features, and non-breaking enhancements.
-Additionally any change to the version of a dependency.
+Additionally, any change to the version of a dependency is a minor version change.
 
 ## Patch versions
 
@@ -19,20 +19,6 @@ Patch versions changes are meant only for bug fixes,
 and will not break compatibility of the current major version.
 A patch release will contain a collection of minor bug fixes, 
 or individual major and security bug fixes, depending on severity.
-A patch release will also include warning of upcoming breaking changes, whenever possible.
-
-# What this means for you
-
-We recommend running the most recent version. Here are our suggestions for managing updates:
-
-* X changes will require some effort to incorporate.
-* Y changes will not require significant effort to incorporate.
-
-  * If you have good unit and integration tests, these changes are generally safe to pick up automatically.
-
-* Z changes will not require any changes to your code. Z changes are intended to be picked up automatically.
-
-  * Good unit and integration tests are always recommended.
 
 # Semantic Commits
 
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -21,8 +21,7 @@ A patch release will contain a collection of minor bug fixes,
 or individual major and security bug fixes, depending on severity.
 A patch release will also include warning of upcoming breaking changes, whenever possible.
 
-What this means for you
-=======================
+# What this means for you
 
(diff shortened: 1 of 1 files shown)


VERSIONING.md
```diff
@@ -0,0 +1,55 @@
...
+## Patch versions
+
+Patch versions changes are meant only for bug fixes,
+and will not break compatibility of the current major version.
+A patch release will contain a collection of minor bug fixes, 
+or individual major and security bug fixes, depending on severity.
+A patch release will also include warning of upcoming breaking changes, whenever possible.
+
+What this means for you
+=======================
```
mattsb42-aws: Needs markdown-style heading notation.

VERSIONING.md
```diff
@@ -0,0 +1,54 @@
...
+## Major versions
+
+Major version changes are significant and expected to break backwards compatibility.
+
+## Minor versions
+
+Minor version changes will not break compatibility between the previous minor versions;
+to do so is a bug.
+Encryption SDK changes will also involve addition of optional features, and non-breaking enhancements.
+Additionally any change to the version of a dependency.
```
lizroth: "Additionally, any change to the version of a dependency is a minor version change."

or

"Minor version changes additionally include any change to the version of a dependency."

I don't think that's a sentence as is. :) Just trying to minimize grok effort for readers.

VERSIONING.md
```diff
@@ -0,0 +1,54 @@
...
+  * If you have good unit and integration tests, these changes are generally safe to pick up automatically.
+
+* Z changes will not require any changes to your code. Z changes are intended to be picked up automatically.
+
+  * Good unit and integration tests are always recommended.
+
+# Semantic Commits
+
+We seek to increase clarity at all levels of the update and releases process.
+We require pull requests adhere to the [Conventional Commits][conventional-commits] spec,
```
lizroth: Have we looked at / tried out any of the automated enforcement tools? What would it take to get some of those running, do we know?

SalusaSecondus:
1. Conventional commits seems like a neat idea and depending on how it works here may be worth rolling out more widely
2. The other ESDKs don't follow semantic versioning and so if this does, it will be different. See the [java VERSIONING file](https://github.com/aws/aws-encryption-sdk-java/blob/master/VERSIONING.rst) as an example. Minor changes are permitted backwards incompatible changes which can be:
  a. Usually be ignored because they are rarely encountered/edge-case, and
  b. Can be automatically picked up by any system with a reasonable set of unit-/integration-tests, and
  c. (If not extremely urgent) are warned with deprecation notices in a prior Minor/Patch release

VERSIONING.md
seebees: I am working on that.  exploring https://github.com/probot/semantic-pull-requests

lizroth: Love it!

VERSIONING.md
```diff
@@ -11,28 +11,14 @@ Major version changes are significant and expected to break backwards compatibil
...
-We recommend running the most recent version. Here are our suggestions for managing updates:
-
-* X changes will require some effort to incorporate.
-* Y changes will not require significant effort to incorporate.
-
-  * If you have good unit and integration tests, these changes are generally safe to pick up automatically.
-
-* Z changes will not require any changes to your code. Z changes are intended to be picked up automatically.
-
-  * Good unit and integration tests are always recommended.
```
lizroth: I think we could have left these as expressions of our intent (especially the bit about if we know we will need to make breaking changes soon, we will make an effort to warn you), but I don't think removing them is a problem either, given that we're stating our use of CC/SemVer.

SalusaSecondus:
To be clear. My concerns aside, I have no objections to us moving forward with this proposed versioning scheme.

PR #227, opened 2019-10-14
fix: eval in portableTimingSafeEqual

seebees:
`eval` was added to help ensure that `portableTimingSafeEqual`
is not optimized into a non-constant time function.
However, the Content Security Policy `'unsafe-eval'` will flag this.

It is better to remove this and risk such an optimization,
then to force customers to weaken the Content Security Policy.

# Check any applicable:

### DIFF
--- a/modules/material-management/src/cryptographic_material.ts
+++ b/modules/material-management/src/cryptographic_material.ts
@@ -86,7 +86,7 @@ const timingSafeEqual: (a: Uint8Array, b: Uint8Array) => boolean = (function () 
      * The eval below is commented out
      * because if a browser is using a Content Security Policy with `'unsafe-eval'`
      * it would fail on this eval.
-     * The value in attempting to ensure this function is not optimized,
+     * The value in attempting to ensure that this function is not optimized
      * is not worth the cost of making customers allow `'unsafe-eval'`.
      * If you copy this function for your own use, make sure to educate yourself.
      * Side channel attacks are pernicious and subtle.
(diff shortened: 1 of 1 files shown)


modules/material-management/src/cryptographic_material.ts
```diff
@@ -82,14 +82,19 @@ const timingSafeEqual: (a: Uint8Array, b: Uint8Array) => boolean = (function ()
...
-      * Side channel attacks are pernicious and subtle.
-      */
-    eval('') // eslint-disable-line no-eval
+     * Adding `eval` could prevent the optimization, but this is no grantee.
+     * The eval below is commented out,
+     * because if a browser is using a Content Security Policy with `'unsafe-eval'`
+     * it would fail on this eval.
+     * The value in attempting to ensure this function is not optimized,
+     * is not worth the cost of making customers to alow `'unsafe-eval'`.
+     * If you copy this function for your own use, make sure to educate yourself.
```
mattsb42-aws: Do we have a reference for such a person to read up on? Just saying "educate yourself" comes across as dismissive, and I know that was not your intent.

modules/material-management/src/cryptographic_material.ts
```diff
@@ -82,14 +82,19 @@ const timingSafeEqual: (a: Uint8Array, b: Uint8Array) => boolean = (function ()
...
+     * Side channel attacks are pernicious and subtle.
+     */
+    // eval('') // eslint-disable-line no-eval
     /* Check for early return (Postcondition) UNTESTED: Size is well-know information.
```
mattsb42-aws:
```suggestion
    /* Check for early return (Postcondition) UNTESTED: Size is well-know information
```

modules/material-management/src/cryptographic_material.ts
seebees: This is a good point.  https://codahale.com/a-lesson-in-timing-attacks/ is linked above.  Perhaps adding `please`?  I was hoping that next sentence might mitigate some of this.

mattsb42-aws: Maybe "please review the above link before use"?

PR #239, opened 2020-01-09
Zero byte operations

### LINKED ISSUE #237: Raw aes keyrings in a browse not processing  zero byte AES-GCM operations.
A few browsers, notably Safari (WebKit), do not properly handle zero byte AES-GCM operations.
In this case a fallback is required.
However for raw AES keyrings the master key is not properly handling this case.

Fix the import and use of the master key.
Add tests.

cc @khadir-syed moving the discussion to here to track the work.

seebees:
Resolves #237 

Several changes bundled together:

* Change the references to `CryptoKey` to `AwsEsdkJsCryptoKey`
* Raw AES Keyring master keys were not correctly using the non-zero backend
* The WebCrypto decrypt API combines the data and tag together, so zero bytes of data is not exactly zero.
* Small, remove redundant check to `'algorithm'`
* For MSRCrypto update the `algorithm` argument for `importKey`

# Check any applicable:


modules/example-browser/src/fallback.ts
```diff
@@ -0,0 +1,118 @@
...
+  decrypt,
+  synchronousRandomValues,
+  configureFallback,
+  AlgorithmSuiteIdentifier
+} from '@aws-crypto/client-browser'
+import { toBase64 } from '@aws-sdk/util-base64-browser'
+
+// @ts-ignore
+import { subtle } from 'msrcrypto'
+configureFallback(subtle)
```
mattsb42-aws: Comment please, highlighting that this is where we configure a fallback and when it will be used.

modules/example-browser/src/fallback.ts
```diff
@@ -0,0 +1,118 @@
...
+ * or in the "license" file accompanying this file. This file is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* This is a simple example of using a raw AES keyring
+ * to encrypt and decrypt using the AWS Encryption SDK for Javascript
+ * in a browser.
+ */
```
mattsb42-aws: Specifically, this example shows how to do that while providing a fallback to use in the event of some missing WebCrypto features. That should be mentioned in this intro.

seebees: fixed

modules/example-browser/src/fallback.ts
seebees: added, good catch.

PR #276, opened 2020-03-18
chore: add verdaccio publishing tests to CI

### LINKED ISSUE #123: add Verdaccio publish/integration-test step to CI
To avoid issues with accidentally creating invalid/incomplete package artifacts, let's add a CI stage that spins up a Verdaccio server, publishes all packages to that server, then goes into a temp directory, installs everything from the Verdaccio server, and runs all integration tests.

seebees:
resolves #123

To validate that everything always works,
running the integration tests directly from npm
will replicate exactly how customers would use this.
It will also verify that all dependencies of the clients
can be satisfied from npm.
Finally, it exercises the publishing process
to make sure that each release could be published automatically.

# Check any applicable:

### DIFF
--- a/util/local_verdaccio_publish
+++ b/util/local_verdaccio_publish
@@ -1,0 +1,55 @@
+#!/usr/bin/env node
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// This is done in a node script for 2 reasons
+// 1. Potability: I will need to run this in windows
+// 2. Errors. If the main command fails, the script needs to clean up
+// but and exit with this error.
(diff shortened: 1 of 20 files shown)


util/local_verdaccio_publish
```diff
@@ -0,0 +1,32 @@
...
+  --no-git-tag-version \
+  --no-push \
+  --no-git-reset \
+  --ignore-scripts \
+  --preid ci \
+  --no-verify-access
+
+# The above command will make some modifications,
+# Roll them back
+# Idealy, we would find a way to not have to do this
```
scottarc: Finding a way to not have to do this might require an upstream change to lerna.

seebees: I agree

PR #281, opened 2020-03-26
fix: The final frame can not be larger than the Frame Length

seebees:
The final frame content length is stored in the body header.
This means that while it should be related to the frame length in the header,
it is not required.

This enforces this requirement.
As well as adding a test and test vector for a zero length final frame.

# Check any applicable:


modules/serialize/src/decode_body_header.ts
```diff
@@ -155,6 +155,8 @@ export function decodeFinalFrameBodyHeader (buffer: Uint8Array, headerInfo: Head
   needs(sequenceNumber > 0, 'Malformed sequenceNumber.')
   const iv = buffer.slice(readPos += 4, readPos += ivLength)
   const contentLength = dataView.getUint32(readPos)
+  /* Postcondition: The final frame should not be able to exceed the frameLength. */
```
mattsb42-aws:
```suggestion
  /* Postcondition: The final frame MUST NOT exceed the frameLength. */
```

modules/serialize/test/decode_body_header.test.ts
```diff
@@ -403,6 +427,24 @@ describe('decodeFinalFrameBodyHeader', () => {
       expect(test).to.eql(false)
     }
   })
+
+  it('Postcondition: The final frame should not be able to exceed the frameLength.', () => {
```
mattsb42-aws:
```suggestion
  it('Postcondition: The final frame MUST NOT exceed the frameLength.', () => {
```

seebees:
@mattsb42-aws That is the edge of the edge.
But... I've been supper happy with all the small component testing in `serialize` so...

modules/encrypt-node/src/framed_encrypt_stream.ts
```diff
@@ -190,11 +190,18 @@ type EncryptFrameInput = {
...
   const { sequenceNumber, contentLength, content } = pendingFrame
-  const frameIv = serialize.frameIv(messageHeader.headerIvLength, sequenceNumber)
+  const { frameLength, contentType, messageId, headerIvLength } = messageHeader
+  /* Precondition: The content length MUST correlate with the frameLength.
+   * In the case of a regular frame,
+   * the content length must strictly equal the frame length.
+   * In the case of the final frame,
+   * it MUST not be larger than the frame length.
+   */
+  needs(frameLength === contentLength || (isFinalFrame && frameLength >= contentLength), 'Final frame length exceeds frame length.')
```
mattsb42-aws: Strictly speaking, this should be two separate checks, no? If something goes terribly wrong on a non-final frame and I get an error that's talking about final frame length, I'm gonna be confused.

seebees: Perhaps.  What about the error message `Content length exceeds frame length`

mattsb42-aws: How about:

```suggestion
  needs(isFinalFrame || frameLength === contentLength, 'Content length does not match frame length.')
  needs(!isFinalFrame || frameLength >= contentLength, 'Final frame length exceeds frame length.')
```

mattsb42-aws: Alternate suggestion:
```suggestion
  if (isFinalFrame){
    needs(frameLength >= contentLength, 'Final frame length exceeds frame length.')
  }else{
    needs(frameLength === contentLength, 'Content length does not match frame length.')
  }
```

seebees: Talking about this, having the exact values in question in the error message should resolve this.

Both the error message AND the condition should clearly express the intent.
Further, if this error is ever hit, knowing if you are a final frame or not will GREATLY simplify the debugging.

PR #285, opened 2020-04-01
Update tests for dependencies

seebees:
This PR is to deal with version updates failing tests.
The tests were passing, Karma was starting to see errors,
both on Typescript compile and on the size of the bundle.

First: Remove mocha import to shrink bundle

Originally, `import 'mocha’` was added to satisfy Typescript.
But this made `webpack` include all of `mocha` in the bundle.
Karma was starting to run out of memory when processing all the test files.

Second: update karma config

The config was in a tenuous state functioning.
Changing the config to the root Typescript config file
that references each module,
and enabling projectReferences make the configuration explicit.
Also adding a log for errors
such as when AWS credentials are not available
makes this error much more obvious.

Third: update dependency versions

This is an explicit update to versions.
Additionally all sub modules package-lock.json files are removed.

When `lerna link` was run, in #242
all modules were linked together at the root.
This means that the root package-lock.json file is responsible
for all sub modules state.
Lerna does not maintain the state on these modules package-lock.json anymore.
So this information is redundant.

Finally, update `local_verdaccio_publish`
to not reset package-lock.json files as they do not exist.

# Check any applicable:


modules/caching-materials-manager-browser/test/caching_materials_manager_browser.test.ts
```diff
@@ -16,7 +16,7 @@
 /* eslint-env mocha */
 
 import { expect } from 'chai'
-import 'mocha'
+// import 'mocha'
```
scottarc, line 19: In most other files, this import was deleted. In this file, it was commented out.

modules/web-crypto-backend/test/backend-factory.test.ts
```diff
@@ -18,7 +18,7 @@
 import * as chai from 'chai'
 import chaiAsPromised from 'chai-as-promised'
 import sinon from 'sinon'
-import 'mocha'
+// import 'mocha'
```
scottarc, line 21: This was also commented out instead of deleted.

PR #316, opened 2020-04-24
fix: nonFramed maximum content length

seebees:
nonFramed messages are encrypted under a single operation.
The maximum number of bytes for a single AES-GCM "operation."
is 2 ** 36 - 32
This is related to the GHASH block size,
and can be thought of as the maximum bytes
that can be encrypted with a single IV.
The AWS Encryption SDK for Javascript
does not support non-framed encrypt
https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/data-format/message-body.md#non-framed-data

# Check any applicable:

### DIFF
--- a/modules/serialize/src/decode_body_header.ts
+++ b/modules/serialize/src/decode_body_header.ts
@@ -248,7 +248,7 @@ export function decodeNonFrameBodyHeader(
    * https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/data-format/message-body.md#encrypted-content-length
    */
   needs(
-    Maximum.BYTES_PER_AES_GCM_NONCE >= contentLength,
+    Maximum.BYTES_PER_AES_GCM_NONCE > contentLength,
     'Content length out of bounds.'
   )
   return {
--- a/modules/serialize/src/decode_body_header.ts
+++ b/modules/serialize/src/decode_body_header.ts
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import BN from 'bn.js'
-import { ContentType, SequenceIdentifier } from './identifiers'
+import { ContentType, SequenceIdentifier, Maximum } from './identifiers'
 import {
   HeaderInfo,
   BodyHeader,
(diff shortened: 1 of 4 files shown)


mattsb42-aws:
>  and can be thought of as the maximum bytes
that can be encrypted with a single IV.

nit: single key/IV pair

modules/serialize/src/identifiers.ts
```diff
@@ -74,6 +74,18 @@ export enum Maximum {
    * this value is set to 2 ** 53 - 1.
    */
   BYTES_PER_MESSAGE = 2 ** 53 - 1,
+  /* Maximum number of bytes for a single AES-GCM "operation."
+   * This is related to the GHASH block size,
+   * and can be thought of as the maximum bytes
+   * that can be encrypted with a single IV.
```
mattsb42-aws, line 80:
```suggestion
   * that can be encrypted with a single key/IV pair.
```

modules/serialize/src/decode_body_header.ts
```diff
@@ -244,6 +244,13 @@ export function decodeNonFrameBodyHeader(
   // This will throw if the number is larger than Number.MAX_SAFE_INTEGER.
   // i.e. a 53 bit number
   const contentLength = contentLengthBN.toNumber()
+  /* Postcondition: Non-framed content length MUST NOT exceed AES-GCM safe limits.
+   * https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/data-format/message-body.md#encrypted-content-length
+   */
+  needs(
+    Maximum.BYTES_PER_AES_GCM_NONCE >= contentLength,
```
mattsb42-aws, line 251:
```suggestion
    Maximum.BYTES_PER_AES_GCM_NONCE > contentLength,
```

The maximum is permissible. There may have been some confusion between the two checks in C?

https://github.com/aws/aws-encryption-sdk-c/blob/eb99852908f21f8ab7bf4567cd31a31a934d63a5/source/framefmt.c#L312

https://github.com/aws/aws-encryption-sdk-c/blob/eb99852908f21f8ab7bf4567cd31a31a934d63a5/source/framefmt.c#L208

I don't remember why the two checks are different or if this was intentional.

The spec allows max[1], Python allows max2], Java..I'm not finding a check, but I might be missing something.

[1] https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/data-format/message-body.md#encrypted-content-length
[2] https://github.com/aws/aws-encryption-sdk-python/blob/00c7e1dbfd0aade7ccb61e5e5735e26a8a239c38/src/aws_encryption_sdk/streaming_client.py#L408-L411

Issue #273, opened 2020-03-13
[keyring] Fix KMS keyring behavior on encrypt


mattsb42-aws:
Once https://github.com/awslabs/aws-encryption-sdk-specification/issues/73 is resolved, we need to change the behavior of the KMS keyring to match the new spec definition.

seebees:
This is resolved by #408.

PR #317, opened 2020-04-24
fix!: The AWS Kms keyring use of clientProvider

seebees:
On encrypt if an AWS KMS client can not be obtained
Encrypt should HALT.

There MUST be a client for every KeyId
When a user configures a AWS KMS keyring with IDs
the intent is to be able to independently decrypt
with ANY of these IDs.
See: https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/framework/kms-keyring.md#onencrypt-goal
Without a client it is
impossible to encrypt under the given CMK.

BREAKING CHANGE:

The previous behavior was to
allow the client supplier to modify
the AWS KMS KeyId intent.
This resulted in a successful
encrypt but WITHOUT the filtered KeyIds

# Check any applicable:

### DIFF
--- a/modules/kms-keyring/src/helpers.ts
+++ b/modules/kms-keyring/src/helpers.ts
@@ -51,12 +51,22 @@ export async function encrypt<Client extends AwsEsdkKMSInterface>(
   KeyId: string,
   EncryptionContext: EncryptionContext,
   GrantTokens?: string[]
-): Promise<RequiredEncryptResponse | false> {
+): Promise<RequiredEncryptResponse> {
   const region = regionFromKmsKeyArn(KeyId)
   const client = clientProvider(region)
 
-  /* Check for early return (Postcondition): clientProvider did not return a client for encrypt. */
-  if (!client) return false
+  /* Postcondition: There MUST be a client for every KeyId
+   * When a user configures a KMS keyring with IDs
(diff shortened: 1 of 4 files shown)


seebees:
This resolves #273

modules/kms-keyring/src/helpers.ts
```diff
@@ -51,12 +51,22 @@ export async function encrypt<Client extends AwsEsdkKMSInterface>(
...
+  /* Postcondition: There MUST be a client for every KeyId
+   * When a user configures a KMS keyring with IDs
+   * the intent is to be able to independently decrypt
+   * with ANY of these IDs.
+   * See: https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/framework/kms-keyring.md#onencrypt-goal
+   * Without a client it is
+   * impossible to encrypt under the given CMK.
+   */
+  if (!client)
+    throw new Error(
```
mattsb42-aws, line 67: Do we have any more granular errors that we can throw? I forget if we did anything with this before elsewhere in this project.

seebees: We do not have any subclassed errors or error codes,
so no :(

seebees:
Moving this to spec v1

PR #319, opened 2020-04-25
fix: Better messages when no keyring succeeded to decrypt

seebees:
Related #152 

If no keyrings attempt to decrypt any encrypted data keys,
then the message can not be decrypted.
The code attempted to enforce this,
by retrieving the unencrypted data key in node.

There were two issues here

1. The check ensure the validity of the materials,
itself threw an error.
1. Had this check succeeded, the error message
`'Unencrypted data key is invalid.’` is not incredibly more helpful than
'unencryptedDataKey has not been set'

The error message has been updated,
and the tests have been updated
to verify _this_ error message.

On a related note
https://github.com/awslabs/aws-encryption-sdk-specification/issues/97
starts to explore some additional possibilities.
The fullness of this issue is not only in failure,
but success can also have similar issues.

# Check any applicable:


seebees:
moving to spec v1

PR #320, opened 2020-04-25
fix: Missing errors in raw keyrings

seebees:
Related #152 

On decrypt, error messages can be passed over.
However, if the keyring attempts to decrypt,
and fails how does anyone know what happend?

This is a similar solution to #212

# Check any applicable:


seebees:
moving this to spec v1

PR #395, opened 2020-08-04
chore: update dependancies

seebees:
Dependabot does not play well with lerna and npm.

Also, update a test file to match the typescript version.

# Check any applicable:


seebees:
No @alex-chew the changes in source were easy to catch, but the tests did not fail until the update was pushed.  When building the other PR, I should have applied the updates and tested again, just to make sure :(

Issue #388, opened 2020-07-20
Add maxHeaderSize Decrypt Input


lavaleri:
See https://github.com/awslabs/aws-encryption-sdk-specification/issues/182

seebees:
Our plan is to track compliance to the specification with a new mechanism

PR #401, opened 2020-08-05
feat: Add maxHeaderSize

seebees:
Resolves: #388

Add maxHeaderSize feature to control
the maximum size of a header that is allowed.

Processing ESDK messages
of unknown size introduces a risk
of highly-bounded resource consumption
due to variable length fields
in the message format.
This is bad for users
that need to decrypt such messages,
especially for such users
that care about
the memory usage of their systems.

For more details see:
https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/proposals/2020-06-26_decrypt-max-header-size-max-body-size/proposal.md

# Check any applicable:

### DIFF
--- a/modules/decrypt-node/src/parse_header_stream.ts
+++ b/modules/decrypt-node/src/parse_header_stream.ts
@@ -22,12 +22,20 @@ const PortableTransformWithType = PortableTransform as new (
 interface HeaderState {
   buffer: Buffer
   headerParsed: boolean
+  maxHeaderSize?: number
+}
+
+export interface ParseHeaderStreamOptions {
+  maxHeaderSize?: number
 }
 
 export class ParseHeaderStream extends PortableTransformWithType {
   private materialsManager!: NodeMaterialsManager
   private _headerState: HeaderState
-  constructor(cmm: NodeMaterialsManager) {
+  constructor(
+    cmm: NodeMaterialsManager,
+    { maxHeaderSize }: ParseHeaderStreamOptions = {}
+  ) {
     super()
     Object.defineProperty(this, 'materialsManager', {
       value: cmm,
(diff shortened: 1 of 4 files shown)


modules/decrypt-node/test/parse_header_stream.test.ts
```diff
@@ -39,15 +42,38 @@ describe('ParseHeaderStream', () => {
       await testStream(cmm, data.slice(0, i))
     }
   })
+
+  it('Precondition: If maxHeaderSize was set I can not buffer a header larger than maxHeaderSize.', async () => {
```
mattsb42-aws, line 46: Just to make sure, let's also add a "equal to length" test case too.

modules/decrypt-node/src/parse_header_stream.ts
```diff
@@ -36,14 +44,34 @@ export class ParseHeaderStream extends PortableTransformWithType {
...
+    const { buffer, maxHeaderSize } = this._headerState
     const headerBuffer = Buffer.concat([buffer, chunk])
-    const headerInfo = deserialize.deserializeMessageHeader(headerBuffer)
+
+    /* maxHeaderSize is the maximum
+     * size we MUST attempt
+     * to deserialize.
+     */
+    const headerInfo = deserialize.deserializeMessageHeader(
+      maxHeaderSize && headerBuffer.byteLength > maxHeaderSize
```
mattsb42-aws, line 60: Let's consolidate this and line 74 into a single resolution of this logic check that we reference both places. That should make this simpler to read and also less likely to diverge in the future.

modules/decrypt-node/src/parse_header_stream.ts
```diff
@@ -36,14 +44,34 @@ export class ParseHeaderStream extends PortableTransformWithType {
...
+        ? headerBuffer.slice(0, maxHeaderSize)
+        : headerBuffer
```
mattsb42-aws, line 62: Suggestion: rather than passing in a slice sometimes and sometimes the original, what if we always pass in a slice and just change where the end of the slice is?

modules/decrypt-node/test/parse_header_stream.test.ts
seebees: This tests this too.  The `completeHeaderLength` holds the value, and the second for loop goes from equality, through the rest of the message.  Adding comments

modules/decrypt-node/test/parse_header_stream.test.ts
```diff
@@ -39,15 +42,50 @@ describe('ParseHeaderStream', () => {
...
+    }
+
+    /* Picking up from the exact header size
+     * test every maxHeaderSize equal to
+     * or greater than the actual header size
+     * (up to the message size).
+     */
+    for (let i = completeHeaderLength; data.byteLength > i; i++) {
+      await testStream(cmm, data.slice(0, i), { maxHeaderSize: i })
+    }
```
alex-chew, line 77: It's not immediately obvious to me that `data.byteLength > completeHeaderLength` ever holds (even if it currently does), and if it doesn't hold then nothing is tested here. Would it make sense to assert that `data.byteLength > completeHeaderLength`, i.e. that the loop runs at least once?

modules/decrypt-node/test/parse_header_stream.test.ts
```diff
@@ -8,22 +8,41 @@ import chaiAsPromised from 'chai-as-promised'
...
       fixtures.base64CiphertextAlgAes256GcmIv12Tag16HkdfWith4Frames(),
       'base64'
     )
+
+    const headerInfo = deserializeMessageHeader(data)
+    needs(headerInfo, 'No header, test impossible')
+    const completeHeaderLength =
+      headerInfo.rawHeader.byteLength +
+      headerInfo.algorithmSuite.ivLength +
+      headerInfo.algorithmSuite.tagLength / 8
```
alex-chew, line 44: It would be good to explicitly compute `completeHeaderLength` this way in the next text fixture as well (`it('Precondition: If maxHeaderSize was set...')`)

seebees: reasonable, I should fix this elsewhere as well.
But I'll do that in a different PR

Issue #390, opened 2020-07-20
Hold final frame plaintext on Decrypt until signature verifies


lavaleri:
See https://github.com/awslabs/aws-encryption-sdk-specification/issues/184

seebees:
Our plan is to track compliance to the specification with a new mechanism

PR #403, opened 2020-08-06
feat: Verify signature before final frame

seebees:
Resolves: #390

Data MUST be authenticated before it is released.
But there is a unique edge case
In the end of signed streams.
If the final frame is sent
but the signature has not been sent
the stream will release the plaintext
because the authTag has been received.
However the signature
has still not been verified.

This could lead to customers processing
data that is not valid
and circumvent one of the values of the signature suites
i.e. someone with decrypt can not forage
a new message.

# Check any applicable:

### DIFF
--- a/modules/decrypt-node/src/verify_stream.ts
+++ b/modules/decrypt-node/src/verify_stream.ts
@@ -252,7 +252,11 @@ export class VerifyStream extends PortableTransformWithType {
     callback()
   }
 
-  _transformSignature = (chunk: Buffer, _enc: string, callback: Function) => {
+  _transformSignature = (
+    chunk: Buffer,
+    _enc: string,
+    callback: (err?: Error) => void
+  ) => {
     const state = this._verifyState
 
     // EOF is an empty buffer, so I need to handle that kind of end.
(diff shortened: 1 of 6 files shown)


modules/decrypt-node/src/verify_stream.ts
```diff
@@ -195,53 +197,78 @@ export class VerifyStream extends PortableTransformWithType {
...
+
+    // EOF is an empty buffer, so I need to handle that kind of end.
+    if (chunk.byteLength === 0) return callback()
+
+    try {
+      const { signatureInfo, finalAuthTag } = state
+      /* Precondition: Only buffer data if the finalAuthTag has been received. */
+      needs(finalAuthTag, 'Malformed state.')
+      /* Precondition: Only buffer data if a signature is expected. */
+      needs(signatureInfo && this._verify && !state.signature, 'To much data')
```
alex-chew, line 253: nit: typo?

modules/decrypt-node/src/verify_stream.ts
```diff
@@ -251,28 +278,36 @@ export class VerifyStream extends PortableTransformWithType {
...
+       */
+      if (this._verify) {
+        /* Precondition: A complete signature is required to verify. */
+        needs(signature, 'Incomplete signature')
+        const { buffer, byteOffset, byteLength } = signature
+        const isVerified = this._verify.awsCryptoVerify(
+          Buffer.from(buffer, byteOffset, byteLength)
+        )
+        /* Postcondition: The signature must be valid. */
+        needs(isVerified, 'Invalid Signature')
```
alex-chew, line 304: nit: s/S/s/

seebees:
```suggestion
        needs(isVerified, 'Invalid signature')
```

alex-chew: Yes, and the corresponding tests should change as well.

Issue #397, opened 2020-08-04
Accept local CMC capacity value of 0


alex-chew:
Follow-up to awslabs/aws-encryption-sdk-specification/pull/166.

Current implementation: https://github.com/aws/aws-encryption-sdk-javascript/blob/20045ca0cae056184ea9cdfdafb83dbc811c22b3/modules/cache-material/src/get_local_cryptographic_materials_cache.ts#L21-L33

The LRU constructor interprets `max: 0` as `max: Inifinity`.

seebees:
Our plan is to track compliance to the specification with a new mechanism

PR #410, opened 2020-08-07
feat: CMC capacity of 0 is a null-cache

seebees:
resolves: #397

An unbounded cache is not cryptographically safe.
A null-cache is useful.

This change makes it very explicit
when customers create an unbounded cache.
And offers a null-cache for testing or debugging.

# Check any applicable:

### DIFF
--- a/modules/cache-material/src/get_local_cryptographic_materials_cache.ts
+++ b/modules/cache-material/src/get_local_cryptographic_materials_cache.ts
@@ -24,12 +24,45 @@ export function getLocalCryptographicMaterialsCache<
   capacity: number,
   proactiveFrequency: number = 1000 * 60
 ): CryptographicMaterialsCache<S> {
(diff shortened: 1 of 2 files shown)


modules/cache-material/src/get_local_cryptographic_materials_cache.ts
```diff
@@ -24,12 +24,45 @@ export function getLocalCryptographicMaterialsCache<
   capacity: number,
   proactiveFrequency: number = 1000 * 60
 ): CryptographicMaterialsCache<S> {
+  /* Precondition: A capacity of 0 is defined as a null-cache.
+   * The underlying `lur-cache` treats `max` === 0
```
alex-chew, line 28:
```suggestion
   * The underlying `lru-cache` treats `max` === 0
```

alex-chew (dismissed):

PR #496, opened 2021-01-05
fix: better dependency injection

seebees:
Sinon is nice, but can not support native ES6 modules.
Because these modules are immutable on import.
Since the backen-factory and the synchronousRandomValues
function both share the same `window` object,
this can be trivially accomplished.
The other advantage is that the test are _much_ simpler.

# Check any applicable:

### DIFF
--- a/modules/web-crypto-backend/test/synchronous_random_values.test.ts
+++ b/modules/web-crypto-backend/test/synchronous_random_values.test.ts
@@ -4,9 +4,8 @@
 /* eslint-env mocha */
 
 import { expect } from 'chai'
+import { generateSynchronousRandomValues } from '../src/synchronous_random_values'
 import { synchronousRandomValues } from '../src/index'
-import sinon from 'sinon'
-import * as browserWindow from '@aws-sdk/util-locate-window'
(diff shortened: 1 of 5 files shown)


modules/web-crypto-backend/test/synchronous_random_values.test.ts
```diff
@@ -4,9 +4,8 @@
 /* eslint-env mocha */
 
 import { expect } from 'chai'
+import { generateSynchronousRandomValues } from '../src/synchronous_random_values'
 import { synchronousRandomValues } from '../src/index'
```
alex-chew, line 8: nitpick: can these can be loaded from the same module?

alex-chew (approved):
Overall LGTM

modules/web-crypto-backend/test/synchronous_random_values.test.ts
seebees: Yes, they can, but I separate them because the one from `index` in the one that is exposed publicly.
Today, they are the same, but they do not need to be, so I lean towards doing it this way...

Issue #507, opened 2021-01-06
KMS Stream Decryption Error


cbeall:
I've been having issues similar to #376 and #423 with decryption streams and S3 where the decryption fails with an error

> Error: Unsupported state or unable to authenticate data
>     at Decipheriv.final (internal/crypto/cipher.js:174:29)
>     at DecipherStream._onAuthTag (/app/aws-perm-test/node_modules/@aws-crypto/decrypt-node/build/main/decipher_stream.js:72:39)

I'm using `@aws-crypto/client-node: 2.0.0` and primarily NodeJS 14.15.4 but have experienced the issues on node 12 and 10 as well.

From what I can tell its not an issue with S3 specifically, but rather just with streams and larger files and maybe I/O latency. I've included some sample code using file streaming instead of S3 that will produce the error in certain circumstances. Its a bit of a contrived example but it should demonstrate the behavior without introducing S3.

```javascript
const { createReadStream, createWriteStream} = require('fs');
const { promisify } = require('util');
const { finished } = require('stream');
const { KmsKeyringNode, buildClient} = require('@aws-crypto/client-node');

const finishedAsync = promisify(finished)

// Required values for test
const kmsKeyId = process.env.KMS_KEY;
const inputFile = process.env.INPUT_FILENAME;
const outputFile = process.env.OUTPUT_FILENAME;

const encryptKeyring = new KmsKeyringNode({
  generatorKeyId: kmsKeyId,
});

const { encryptStream, decryptStream } = buildClient();
const encryptedFileName = 'file.encrypted';

(async function() {
  const step1Stream = createReadStream(inputFile)
    .pipe(encryptStream(encryptKeyring))
    .pipe(createWriteStream(encryptedFileName));

  await finishedAsync(step1Stream);

  const step2Stream = createReadStream(encryptedFileName)
    .pipe(decryptStream(new KmsKeyringNode({ discovery: true })))
    .once('error', (err) => {
      console.error(err);
    })
    .pipe(createWriteStream(outputFile))
    

  await finishedAsync(step2Stream);

  console.log(`Process complete`);
}())
```

On my machine, this code works with small files (under 1MB) consistently. Files around 10MB seem to work sometimes but not others and any file over 100MB always fails. And if you comment out the pipe with the encryptStream and decryptStream, the remaining code works for files of any size.

Any help would be appreciated. Thank you.

seebees:
This is an excellent report, thanks!
I’ve been able to reproduce this
[...]

PR #509, opened 2021-01-08
fix: Boundary condition error in VerifyStream

seebees:
Fixes: #507

Tests exist that push a single byte at a time.
This will hit every boundary condition.
However this is not complete enough.
State can be reset as the code transitions
from one boundary condition to another.

The prime example of this problem
the `VerifyStream`.
It has 4 boundary in the `_transform` function.
The body header, the body, the auth tag, and signature.
By sending 1 byte
as the code transitions from the body header
to the body
the code handling the exiting
of the body header has no way to interact
with the body,
because the 1 byte of data
completely isolates these code branches.

In this case the code was buffering
the body header in `frameBuffer`
but when exiting this section
and passing control to
the actual content section,
it used the `chunk` to form the tail.

When passing a single byte,
this tail would _always_ be empty,
because it would have been incorporated
into the frame header.

Adding tests to span
the boundary conditions
to verify both this fix
and that there are not other lurking.

# Check any applicable:

### DIFF
--- a/modules/decrypt-node/test/decrypt.test.ts
+++ b/modules/decrypt-node/test/decrypt.test.ts
@@ -70,7 +70,7 @@ describe('decrypt', () => {
      * but it will never span any boundary conditions.
      * The prime example of this is
      * the `VerifyStream`.
-     * It has 4 boundary in the `_transform` function.
+     * It has 4 byte boundary in the `_transform` function.
      * The body header, the body, the auth tag, and signature.
      * By sending 1 byte
      * as the code transitions from the body header
--- a/modules/decrypt-node/test/decrypt.test.ts
+++ b/modules/decrypt-node/test/decrypt.test.ts
@@ -59,33 +59,86 @@ describe('decrypt', () => {
       fixtures.base64CiphertextAlgAes256GcmIv12Tag16HkdfSha384EcdsaP384(),
       'base64'
     )
-    const i = ciphertext.values()
-    const ciphertextStream = from(
-      (
-        _: number,
-        next: (err: Error | null, chunk: Uint8Array | null) => void
-      ) => {
-        /* Pushing 1 byte at time is the most annoying thing.
-         * This is done intentionally to hit _every_ boundary condition.
-         */
-        const { value, done } = i.next()
-        if (done) return next(null, null)
-        next(null, new Uint8Array([value]))
-      }
+
+    /* Pushing 1 byte at time is annoying.
+     * But there can be state that can get reset
+     * from a 1 byte push when the buffering
+     * is internal to the stream.
+     * So while 1 byte will hit the beginning
+     * boundary conditions,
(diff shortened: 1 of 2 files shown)


modules/decrypt-node/test/decrypt.test.ts
```diff
@@ -59,33 +59,86 @@ describe('decrypt', () => {
...
+     * But there can be state that can get reset
+     * from a 1 byte push when the buffering
+     * is internal to the stream.
+     * So while 1 byte will hit the beginning
+     * boundary conditions,
+     * and the exiting boundary conditions,
+     * but it will never span any boundary conditions.
+     * The prime example of this is
+     * the `VerifyStream`.
+     * It has 4 boundary in the `_transform` function.
```
robin-aws, line 73:
```suggestion
     * It has 4 byte boundary in the `_transform` function.
```

?

modules/decrypt-node/test/decrypt.test.ts
```diff
@@ -59,33 +59,86 @@ describe('decrypt', () => {
...
+     * Doing this randomly is a bad idea.
+     * Tests that only fail _sometimes_
+     * are bad tests.
+     * It is too easy to try again,
+     * and when everything passes just move on.
+     * And again, doing every permutation
+     * is too expensive at this time.
+     */
+    const results = await Promise.all(
+      [
```
robin-aws, line 104: nit: there must be a more compact way to get this value :)

modules/decrypt-node/test/decrypt.test.ts
```diff
@@ -59,33 +59,86 @@ describe('decrypt', () => {
...
     )
 
-    expect(messageHeader.suiteId).to.equal(
-      AlgorithmSuiteIdentifier.ALG_AES256_GCM_IV12_TAG16_HKDF_SHA384_ECDSA_P384
-    )
-    expect(messageHeader.encryptionContext).to.deep.equal(
-      fixtures.encryptionContext()
-    )
-    expect(test.toString('base64')).to.equal(fixtures.base64Plaintext())
+    results.map(({ plaintext: test, messageHeader }) => {
```
robin-aws, line 133: Do we have a good option for more direct support for parameterized tests? It'll really suck to debug which size triggered a test failure if there's no output to tell which size was being used.

robin-aws (changes requested):

modules/decrypt-node/test/decrypt.test.ts
seebees: Not really no :(

modules/decrypt-node/test/decrypt.test.ts
seebees: This is actually in response to your question below.
Since there is no nice parameterized test,
it is easiest to just comment out the ones you don't care about
to run the test and debug the issue...

PR #634, opened 2021-06-29
fix: der2raw sLength is s byte length

seebees:
Raw signature (also see ISO/IEC 7816-8 / IEEE P1363)
are the static size concatenation of the r and s value.
These values MUST be 0 padded to the correct length.

This fixes the occasional `Invalid Signature` in browsers.
This impacts ~1% of messages.
These messages are valid and can now be decrypted.

# Check any applicable:


modules/serialize/test/ecdsa_signatures.test.ts
```diff
@@ -347,1240 +165,206 @@ function test() {
...
+  194, 181, 56, 20, 193, 213, 40, 112, 59, 13, 254, 55, 177, 231, 189,
+  128, 71, 241, 232, 224, 60, 113, 203, 248, 143, 34, 63, 98, 221, 156,
+  143, 58, 106, 169, 169, 63, 126, 103, 145, 63, 246, 255, 32, 74, 11, 197,
+  255, 13, 244, 105, 188, 157, 210, 200, 36, 140, 218, 1, 115, 99, 255, 212,
+  71, 156, 38,
+])
+
+// Suite AlgorithmSuiteIdentifier.ALG_AES256_GCM_IV12_TAG16_HKDF_SHA512_COMMIT_KEY_ECDSA_P384
+// prettier-ignore
+const derSigOnlySPadded = new Uint8Array([
```
texastony, line 258: **Only** concerns me, as it raises confusion. Given all I know, I assume that *derSigOnlySPadded* is different from *derSigSPadded*, and the difference is that `r = 0`, is that right? If not, what is the difference?

modules/serialize/test/ecdsa_signatures.test.ts
```diff
@@ -1620,24 +458,57 @@ describe('raw2der', () => {
...
+      expect(raw2der(rawSigSPadded, suite)).to.deep.equal(derSigSPadded)
+    })
+
+    it('Original vector R and S padded', () => {
+      expect(raw2der(rawSigBothSandRPadded, suite)).to.deep.equal(
+        derSigBothSandRPadded
+      )
+    })
+
+    it('New vector S only padded', () => {
```
texastony, line 488: Very minor: I hate/despise *New* in a repo. It is New today, it is old tomorrow. Is there some other adjective we can use that possible addresses the **Only** comment above?

texastony (changes requested):
It's probably because I am foolish, but if we could address some of the adjectives we are using, this will be a better read.

modules/serialize/test/ecdsa_signatures.test.ts
seebees: I don't like to delete vectors.  How about numbers? `derSigSPadded` `derSigSPadded1` derSigSPadded2`?

texastony (approved):
Extra Groovy!

PR #644, opened 2021-07-06
chore: Add support policy

farleyb-amazon:
*Description of changes:*

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #645, opened 2021-07-07
Deprecate node 8 and 10

seebees:
Related to #644

# Check any applicable:


modules/material-management/src/types.ts
```diff
@@ -113,3 +121,38 @@ export interface ClientOptions {
...
+  dq?: string
+  e?: string
+  k?: string
+  kty?: string
+  n?: string
+  p?: string
+  q?: string
+  qi?: string
+  x?: string
+  y?: string
```
farleyb-amazon, line 138: Can we add comments explaining these? The variable names are opaque to me.

seebees: This is a copy of the `JsonWebKey` type. They are not my invention.

/* These are copies of the v12 Node.js types.
 * I copied them here to avoid exporting v12 types
 * and forcing consumers to install/use v12 in their projects.

is the relevant comment.
It is no longer 12, and now 14 as well as some DOM types.

farleyb-amazon (approved):

Issue #722, opened 2021-08-31
Compile errors @aws-crypto/material-management with @types/node >= 16.0.2


cernytomas:
### Problem:

Cannot build Typescript project using @aws-crypto/material-management and @types/node >= 16.0.2

```
node_modules/@aws-crypto/material-management-node/build/main/src/material_helpers.d.ts:3:10 - error TS2724: '"crypto"' has no exported member named 'Signer'. Did you mean 'Sign'?

3 import { Signer, Verify, createCipheriv, createDecipheriv } from 'crypto';
           ~~~~~~

  node_modules/@types/node/crypto.d.ts:1265:11
    1265     class Sign extends stream.Writable {
                   ~~~~
    'Sign' is declared here.
```

### Solution:

This issue is due to wrong type definition of crypto `Sign` class (previously `Signer`) fixed in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/54394 
released with @types/node@16.0.2

Update @types/node to version 16.0.2 or higher
Rename `Signer` to `Sign`

This solution will be breaking change for developers using @types/node < 16.0.2

tomasstrejcek:
This can probably be closed then? :)

texastony:
Howdy @sandfox , @tomasstrejcek , @cernytomas 

The latest release of the [ESDK-JavaScript](https://github.com/aws/aws-encryption-sdk-javascript/releases/tag/v3.1.0) supports Node 16. 
If there are still problems with using our library with `@types/node >= 16`, please re-open or create a new issue.

Much Obliged,
AWS Crypto Tools

PR #723, opened 2021-08-31
fix: Update @types/node to 16.7.9

cernytomas:
Issue #722 

*Description of changes:*

Updated @types/node to 16.7.8
Renamed `Signer` to `Sign`

# Check any applicable:

### DIFF
--- a/modules/material-management-node/src/material_helpers.ts
+++ b/modules/material-management-node/src/material_helpers.ts
@@ -11,7 +11,7 @@ import {
   NodeHash,
 } from '@aws-crypto/material-management'
 import {
-  Signer,
+  Sign,
   Verify,
   createCipheriv,
   createDecipheriv,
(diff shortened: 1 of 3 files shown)


josecorella (approved):
Hello 👋!

Thank you for contributing to the AWS Encryption SDK JavaScript.
Your changes have been approved!

corrjo:
After this update I'm getting 

```
node_modules/@aws-crypto/material-management-node/build/main/src/material_helpers.d.ts:3:10 - error TS2724: '"crypto"' has no exported member named 'Sign'. Did you mean 'sign'?

3 import { Sign, Verify, createCipheriv, createDecipheriv } from 'crypto';
           ~~~~

  node_modules/@types/node/crypto.d.ts:1159:14
    1159     function sign(
                      ~~~~
    'sign' is declared here.
```

corrjo:
I see, @types/node 14 vs 16 issue. 

I know node 16 is becoming lts in a few weeks, but pushing out a breaking change while the active LTS is 14, and then releasing it as a patch kind of sucks.

texastony:
Howdy @corrjo,

Thank you for bringing this to our attention.
We will issue a rollback.

Our goal is to ensure that our dependencies
are not public in ways that break our customer's code.

Much Obliged,
AWS Crypto Tools

tomasstrejcek:
what a pain :) maybe you could have directly release new minor version supporting new version because both of those are breaking changes, which have no place in a crucial modules like these? :(

PR #741, opened 2021-09-16
feat(node): support node v16

texastony:
*Issue #, if available:* #740, #739, #738, #737, #731, #716, #714, #710, #642

*Description of changes:* Update dependencies.

# Check any applicable:


modules/integration-node/src/decrypt_materials_manager_node.ts
```diff
@@ -2,29 +2,29 @@
...
   KmsKeyInfo,
-  KmsMrkAwareKeyInfo,
   KmsMrkAwareDiscoveryKeyInfo,
+  KmsMrkAwareKeyInfo,
   RSAKey,
-  AESKey,
-  KMSKey,
-  KeyInfoTuple,
-  buildGetKeyring,
+  RsaKeyInfo,
```
texastony, line 27: I apologize for this file. When I used my IDE to move `NotSupported`, it re-ordered all the imports.

There are no changes to the imports at all. Just their ordering.

modules/kms-keyring/src/kms_keyring.ts
```diff
@@ -259,7 +259,9 @@ export function KmsKeyringClass<
            * If the caller does not have access they may have access
            * through another Keyring.
            */
-          cmkErrors.push(e)
+          if (e instanceof Error) {
+            cmkErrors.push(e)
+          }
```
texastony, line 264: All the changes involving `catch(<var>)... <do something with var>` are caused by [Using `unkown` in Catch Variables](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#using-unknown-in-catch-variables).
There are two alternatives to this change:
1. `--useUnknownInCatchVariables`, which will disable this feature globally.
2. Export an interface that can handle `Error | any | unknown` and have `cmkErrors` use that.

modules/raw-rsa-keyring-browser/src/raw_rsa_keyring_web_crypto.ts
```diff
@@ -193,6 +193,7 @@ export class RawRsaKeyringWebCrypto extends KeyringWebCrypto {
     const { wrappingAlgorithm, format, key } = getImportOptions(publicKey)
     const backend = await getWebCryptoBackend()
     const subtle = getNonZeroByteBackend(backend)
+    // @ts-ignore // TS2769: No overload matches this call
     return subtle.importKey(format, key, wrappingAlgorithm, false, ['wrapKey'])
```
texastony, line 197: I am not smart enough to understand how `format` does not resolve to match an override.

I am open to suggestions to avoid `ts-ignore`.

package.json
```diff
@@ -133,7 +133,7 @@
     "rimraf": "^3.0.2",
     "source-map-support": "^0.5.19",
     "tslib": "^2.3.0",
-    "typescript": "^4.0.2",
+    "typescript": "^4.4.3",
```
texastony, line 136: Oh. I thought that we were going from 4.3.5 to 4.4.3... based on #738.

package-lock.json
```diff
@@ -449,6 +444,27 @@
...
             "yargs": "^16.1.1"
+          },
+          "dependencies": {
+            "y18n": {
+              "version": "5.0.8",
+              "resolved": "https://registry.npmjs.org/y18n/-/y18n-5.0.8.tgz",
+              "integrity": "sha512-0pfFzegeDWJHJIAmTLRP2DwHjdF5s7jo9tuztdQxAhINCdvS+3nGINqPd00AphqJR/0LhANUS6/+7SCb98YOfA=="
+            },
+            "yargs": {
+              "version": "16.2.0",
```
texastony, line 455: This should be 17.0.2. Bother.

package-lock.json
```diff
@@ -941,13 +944,21 @@
         "@aws-crypto/integration-vectors": "file:modules/integration-vectors",
         "@types/got": "^9.6.9",
         "@types/stream-to-promise": "^2.2.0",
-        "@types/yargs": "^15.0.3",
+        "@types/yargs": "^17.0.2",
         "got": "^11.8.0",
         "stream-to-promise": "^3.0.0",
         "tslib": "^2.3.0",
         "yargs": "^17.1.0"
```
texastony, line 951: This should be at least 17.1.1 , maybe 17.0.2 ... Bother

texastony: Wait: Line 16895 gives me hope that all is well.

package-lock.json
texastony: Wait: Line 16895 gives me hope that all is well.

modules/decrypt-node/src/verify_stream.ts
```diff
@@ -284,7 +284,9 @@ export class VerifyStream extends PortableTransformWithType {
       needs(isVerified, 'Invalid Signature')
       callback()
     } catch (e) {
-      callback(e)
+      if (e instanceof Error) {
+        callback(e)
+      }
```
seebees, line 289: This is problematic. JS can throw whatever it likes, but if it did throw, there was still an error. An err-back function does not care what the error is.  This may be a TS is being ... too good at its job?

TL;DR; you can't just swallow an error. Even if that error is super unlikely :(

modules/integration-node/src/cli.ts
```diff
@@ -75,11 +75,10 @@ const cli = yargs
...
 
 ;(async (argv) => {
+  // prettier-ignore
   const {
-    _: [command],
-    tolerateFailures,
-    testName,
-    concurrency,
+    // @ts-ignore // ignoring TS2339
+    _: [command], tolerateFailures, testName, concurrency,
```
seebees, line 81: The yargs types update does not fix this?

modules/integration-vectors/src/index.ts
```diff
@@ -1,6 +1,9 @@
 // Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import { NotSupported } from '@aws-crypto/material-management'
+
 export * from './types'
 export * from './get_decrypt_test_iterator'
 export * from './build_get_keyring'
+export { NotSupported }
```
seebees, line 9: you can just `export { NotSupported } from '@aws-crypto/material-management'`
But why?

modules/kms-keyring/src/kms_keyring.ts
seebees: 2 is the better option. Because it is true the JS can throw anything, and many of the checks that you have added have subtle impacts that we want to avoid. e.g. continuing after specific types of errors :(

package.json
seebees: The `^` in the range means that any minor/patch version is acceptable. e.g. all same major. The package-lock.json has what the current version of the dependencies resolves to. What you are doing is raising up the floor of the acceptable range.

modules/integration-vectors/src/index.ts
texastony: Wait.... I do not know what happened here... I think this is from an earlier effort where I did not move NotSupported... I will remove this.
NotSupported should be tied to Needs every where it is... needed.

modules/integration-node/src/cli.ts
texastony: After very careful testing, I can assure you that yargs updates did not fix this. Any ideas? If I have/find any I will try and get rid of this ignore...

texastony: I will try the change log for `yargs`, maybe there was something interesting...

texastony: `await`!!!!

modules/raw-rsa-keyring-browser/src/raw_rsa_keyring_web_crypto.ts
texastony: I tried to address with b7fbd96, but failed. WebCrypto's method overrides for `importKey` are below. 
`KeyFormat` there aliases to `"jwk" | "pkcs8" | "raw" | "spki"`. Clearly, the return of the ESDK's `getImportOptions` will always conform to a valid override of `importKey`. 

There is probably a way to convince TS that is the case. But I think the return annotation in b7fbd96 is enough of an effort for any developer to recognize that all is well. TS/JS will run the code happily; TS is just forcing us to be crystal clear. I suggest that we leave this as is.

```
importKey(format: "jwk", keyData: JsonWebKey, algorithm: AlgorithmIdentifier | RsaHashedImportParams | EcKeyImportParams | HmacImportParams | AesKeyAlgorithm, extractable: boolean, keyUsages: KeyUsage[]): Promise<CryptoKey>;
importKey(format: Exclude<KeyFormat, "jwk">, keyData: BufferSource, algorithm: AlgorithmIdentifier | RsaHashedImportParams | EcKeyImportParams | HmacImportParams | AesKeyAlgorithm, extractable: boolean, keyUsages: KeyUsage[]): Promise<CryptoKey>;
```

modules/integration-browser/src/integration.encrypt.test.ts
```diff
@@ -60,7 +60,8 @@ function aTest(testName: string, decryptOracle: string) {
       needs(response.ok, `Failed to decrypt: ${toUtf8(body)}`)
       expect(plainText).toEqual(new Uint8Array(body))
     } catch (e) {
-      if (!notSupportedMessages.includes(e.message)) throw e
+      if (e instanceof Error && !notSupportedMessages.includes(e.message))
+        throw e
```
seebees, line 64: I think that this logic is not perfect. If something is not supported it is not an error. However in every other case it _is_ an error.

So I think that `needs` may be a better solution.

e.g.

```
needs(e instanceof Error && notSupportedMessages.includes(e.message))
```

modules/integration-browser/src/testDecryptFixture.ts
```diff
@@ -96,7 +96,10 @@ export async function testNegativeDecryptFixture(
     const cmm = await _getCmm(keyInfos)
     decryptResult = await _decrypt(cmm, cipher)
   } catch (err) {
-    if (notSupportedDecryptMessages.includes(err.message))
+    if (
+      err instanceof Error &&
+      notSupportedDecryptMessages.includes(err.message)
+    )
```
seebees, line 102: same as above :)

modules/material-management/src/types.ts
```diff
@@ -156,3 +156,7 @@ export interface AwsEsdkAsymmetricKeyDetails {
    */
   namedCurve?: string
 }
+
+export interface ErrorPlus {
+  errPlus?: Error | string | any
+}
```
seebees, line 162: I don't understand this name

modules/raw-rsa-keyring-browser/src/raw_rsa_keyring_web_crypto.ts
```diff
@@ -203,19 +204,22 @@ export class RawRsaKeyringWebCrypto extends KeyringWebCrypto {
     const backend = await getWebCryptoBackend()
 
     if (isFullSupportWebCryptoBackend(backend)) {
+      // @ts-ignore // TS2769: No overload matches this call
```
seebees, line 207: If we can't convince TS the types work out, the comment should at least explain why 'No overload matches this call' is not real and what really is going on :)

modules/material-management/src/types.ts
texastony: ErrorPlus, as in Error or More. 

Alternative: Catchable? (as in something that can be caught; like a pokemon.)

modules/raw-rsa-keyring-browser/src/raw_rsa_keyring_web_crypto.ts
texastony: I tried the idea discussed on our call. No luck.
I tried ideas around that. Defining a type, using {}, etc.

The only idea I did not try was explicitly creating a type that was either Json or the other override.
That might work. 
But I think I am putting this down as I move off call... not sure yet.

modules/integration-browser/src/integration.encrypt.test.ts
texastony: I struggled with this for most of yesterday. Did not make any progress...

modules/example-browser/package.json
```diff
@@ -34,12 +34,12 @@
...
     "karma-coverage-istanbul-reporter": "^3.0.3",
     "karma-mocha": "2.0.1",
-    "karma-webpack": "^4.0.2",
+    "karma-webpack": "^5.0.0",
     "ts-loader": "9.2.5",
     "ts-node": "^10.2.1",
     "tslib": "^2.2.0",
     "typescript": "^4.0.2",
-    "webpack": "^4.42.1",
+    "webpack": "^5.42.0",
```
texastony, line 42: NPM 7 required this upgrade.

However, when webpack is compiled, it throws an error. 
Interestingly, I believe the compile error is a warning that `Array.from` is no longer injected into browsers that do not support it... I am not dead sure. I am still wrapping my head about what webpack does.

This does bring webpack in `example-browser` inline with the webpack in `integration-browser`.

texastony:
This build will fail because the build limit has not been increased yet.

modules/material-management/src/needs.ts
```diff
@@ -21,3 +21,13 @@ export function needs(
...
+
+export class NotSupported extends Error {
+  code: string
+
+  constructor(message?: string) {
+    super(message)
+    Object.setPrototypeOf(this, NotSupported.prototype)
+    this.code = 'NOT_SUPPORTED'
+  }
+}
```
seebees, line 33: Why does this error go here?
This is for the integration modules.

modules/raw-rsa-keyring-browser/src/raw_rsa_keyring_web_crypto.ts
```diff
@@ -193,34 +194,46 @@ export class RawRsaKeyringWebCrypto extends KeyringWebCrypto {
     const { wrappingAlgorithm, format, key } = getImportOptions(publicKey)
     const backend = await getWebCryptoBackend()
     const subtle = getNonZeroByteBackend(backend)
+    // @ts-ignore // see TS2769 Note below
```
seebees, line 197: I suggest that rather than ignoring we do something like this

```typescript
async function ImportKeyTypeOverload(
  op: ReturnType<typeof getImportOptions>,
  subtle: SubtleCrypto,
  keyUsages: KeyUsage[]
): Promise<AwsEsdkJsCryptoKey> {
  return op.format == 'jwk'
    ? subtle.importKey(
        op.format,
        op.key,
        op.wrappingAlgorithm,
        false,
        keyUsages
      )
    : subtle.importKey(
        op.format,
        op.key,
        op.wrappingAlgorithm,
        false,
        keyUsages
      )
}
```

I'm going to push a change @texastony , let me know what you think :)

seebees (changes requested):

texastony (commented):
Awesome. I spent hours trying different things, glad it was figured out!

modules/material-management/src/needs.ts
texastony: @seebees It is now used in `modules/raw-rsa-keyring-node/src/oaep_hash_supported.ts`.  Line 35.
This change was from a call we had a long time ago when we were first working on this PR.

modules/raw-rsa-keyring-node/src/oaep_hash_supported.ts
```diff
@@ -32,10 +32,13 @@ export const oaepHashSupported = (function () {
...
-      'Unexpected error testing oaepHash.'
-    )
-    return true
+    if (ex instanceof NotSupported) {
+      needs(
+        ex.code === 'ERR_OSSL_EVP_INVALID_DIGEST',
+        'Unexpected error testing oaepHash.'
+      )
+      return true
+    }
```
texastony, line 41: @seebees this is where `NotSupported` is used in a `src` path instead of a test path.

seebees (approved):
LGTM

seebees (approved):
Good :)

seebees (approved):

PR #866, opened 2022-03-10
fix: browser-encrypt can encrypt 0 bytes

seebees:
Browser encryption now correctly encrypts 0 byte messages.
Added structural tests for both Node.js and browser `encrypt`.

# Check any applicable:

### DIFF
--- a/modules/encrypt-browser/src/encrypt.ts
+++ b/modules/encrypt-browser/src/encrypt.ts
@@ -138,14 +138,15 @@ export async function _encrypt(
     await getSubtleEncrypt(headerIv, header)(new Uint8Array(0))
   )
 
-  const numberOfFrames = Math.ceil(plaintextLength / frameLength)
+  // In the case of plaintextLength == 0 there still needs to be 1 frame.
+  const numberOfFrames = Math.max(Math.ceil(plaintextLength / frameLength), 1)
   /* The final frame has a variable length.
    * The value needs to be known, but should only be calculated once.
    * So I calculate how much of a frame I should have at the end.
    * This value will NEVER be larger than the frameLength.
    */
   const finalFrameLength =
-    frameLength - (numberOfFrames * frameLength - plaintextLength)
+    plaintextLength - Math.floor(plaintextLength / frameLength) * frameLength
   const bodyContent = []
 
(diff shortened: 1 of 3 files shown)


modules/encrypt-browser/src/encrypt.ts
```diff
@@ -138,14 +138,15 @@ export async function _encrypt(
...
+  // In the case of plaintextLength == 0 there still needs to be 1 frame.
+  const numberOfFrames = Math.max(Math.ceil(plaintextLength / frameLength), 1)
   /* The final frame has a variable length.
    * The value needs to be known, but should only be calculated once.
    * So I calculate how much of a frame I should have at the end.
    * This value will NEVER be larger than the frameLength.
    */
   const finalFrameLength =
-    frameLength - (numberOfFrames * frameLength - plaintextLength)
+    plaintextLength - Math.floor(plaintextLength / frameLength) * frameLength
```
alex-chew, line 149: nit: since there are `(numberOfFrames - 1)` regular frames, I find this more obviously correct:
```suggestion
    plaintextLength - (numberOfFrames - 1) * frameLength
```

modules/encrypt-node/test/encrypt.test.ts
```diff
@@ -370,6 +370,34 @@ describe('encrypt structural testing', () => {
     // This will throw if it does not deserialize correctly
     deserializeSignature(footerSection)
   })
+
+  it('can encrypt empty message', async () => {
```
alex-chew, line 374: nit: any reason to check the encryption context and EDKs in the browser version of this test but not the Node version?

alex-chew (dismissed):

modules/encrypt-node/test/encrypt.test.ts
seebees: Just a miss, inc

seebees: to elaborate more, when encrypting 0 bytes, the only part of the message that is relevant _is_ the encryption context :)

alex-chew (dismissed):
LGTM

alex-chew (approved):

PR #880, opened 2022-03-25
refactor: replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

# Check any applicable:

### DIFF
--- a/modules/material-management/src/pem_helpers.ts
+++ b/modules/material-management/src/pem_helpers.ts
@@ -92,7 +92,7 @@ export function chunk64(buff: Buffer) {
   const chunks: string[] = new Array(numChunks)
 
   for (let i = 0, o = 0; i < numChunks; ++i, o += chunkSize) {
-    chunks[i] = str.substr(o, chunkSize)
+    chunks[i] = str.slice(o, o + chunkSize)
   }
 
   return chunks


modules/material-management/src/pem_helpers.ts
```diff
@@ -92,7 +92,7 @@ export function chunk64(buff: Buffer) {
   const chunks: string[] = new Array(numChunks)
 
   for (let i = 0, o = 0; i < numChunks; ++i, o += chunkSize) {
-    chunks[i] = str.substr(o, chunkSize)
```
seebees, line 95: `substr` is start, length and `slice` is start end :)
so `"length vs position".substr(5,3) == "h v"`
but `"length vs position".slice(5,3) == ""`

CommanderRoot: You are absolutely right. I must have read "o" as "0" aka zero. I updated the PR

seebees (approved):
Looks good. Thanks!

PR #946, opened 2022-06-17
chore: add memory profiling in streams examples

josecorella:
*Description of changes:*
Adds a directory with instructions on how to run some memory profiling examples when working with streams.

# Check any applicable:


seebees (approved):
LGTM

stream_mem_stress_test/memory.ts
```diff
@@ -0,0 +1,101 @@
...
+    origin: 'us-west-2',
+}
+
+/**
+ * kmsEncryptStream will use the encryptStream function and create a pipeline to encrypt a stream of data
+ *  from a file and write it to destination `./{filename}.encrypted`
+ * @param filename string of file name you wish to encrypt
+ * @param framesize optional parameter to determine frame size; default is 4096 bytes
+ */
+export async function kmsEncryptStream(filename:string, framesize?:number) {
```
texastony, line 51: Nit: There should be a space b/w the variable and its type:
```suggestion
export async function kmsEncryptStream(filename: string, framesize?: number) {
```

stream_mem_stress_test/memory.ts
```diff
@@ -0,0 +1,101 @@
...
+ * that were created with a committing algorithm suite.
+ * This is the default if you build `buildClient()`.
+ */
+const { encryptStream, decryptUnsignedMessageStream } = buildClient(
+    CommitmentPolicy.REQUIRE_ENCRYPT_REQUIRE_DECRYPT
+)
+
+import { finished } from 'stream'
+import { createReadStream, createWriteStream } from 'fs'
+import { promisify } from 'util'
```
texastony, line 22: Nit: I could be wrong, but I think we normally put `import`s at the top of the file, and the define our constants.

stream_mem_stress_test/index.ts
```diff
@@ -0,0 +1,88 @@
...
+    res.write("Attempt to encrypt large random")
+    kmsDecryptStream('./rand_1gb.txt.encrypted')
+    res.send()
+})
+
+app.listen(3000, () => {
+    console.log("Listening on port 3000");
+})
+
+async function readFile(filename:string, framesize?:number) {
```
texastony, line 82: Nit: Space b/w variable and it's type in signatures:
```suggestion
async function readFile(filename: string, framesize?: number) {
```

stream_mem_stress_test/tsconfig.json
```diff
@@ -0,0 +1,103 @@
...
+  "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig to read more about this file */
+
+    /* Projects */
+    // "incremental": true,                              /* Save .tsbuildinfo files to allow for incremental compilation of projects. */
+    // "composite": true,                                /* Enable constraints that allow a TypeScript project to be used with project references. */
+    // "tsBuildInfoFile": "./.tsbuildinfo",              /* Specify the path to .tsbuildinfo incremental compilation file. */
+    // "disableSourceOfProjectReferenceRedirect": true,  /* Disable preferring source files instead of declaration files when referencing composite projects. */
+    // "disableSolutionSearching": true,                 /* Opt a project out of multi-project reference checking when editing. */
+    // "disableReferencedProjectLoad": true,             /* Reduce the number of projects loaded automatically by TypeScript. */
```
texastony, line 11: Style: Let's get rid off all this commented out stuff...

texastony (changes requested):
Need's a little clean up, but other than that, LGTM

stream_mem_stress_test/memory.ts
josecorella: I was following what we do in the node examples:
https://github.com/aws/aws-encryption-sdk-javascript/blob/master/modules/example-node/src/kms_stream.ts#L4-L27

texastony (approved):
I really like your tsconfig.
Very clean

PR #995, opened 2022-08-25
docs(SupportPolicy): Mark 1.x End-of-Support

texastony:
*Issue #, if available:* Mark 1.x End-of-Support

*Description of changes:*
Update Support Policy, marking:
- 1.x as EOS
- 2.x as Maintenance
- 3.x as GA

# Check any applicable:

### DIFF
--- a/SUPPORT_POLICY.rst
+++ b/SUPPORT_POLICY.rst
@@ -22,11 +22,15 @@ This table describes the current support status of each major version of the AWS
       - Next status
       - Next status date
     * - 1.x
+      - Maintenance
       - End of Support
-      - 
-      - 
+      - 2022-07-08
     * - 2.x
       - General Availability
+      - 
+      - 
+    * - 3.x
+      - General Availability 
       - 
       - 
 
--- a/SUPPORT_POLICY.rst
+++ b/SUPPORT_POLICY.rst
@@ -22,15 +22,15 @@ This table describes the current support status of each major version of the AWS
       - Next status
       - Next status date
     * - 1.x
+      - End of Support
+      - 
+      - 
+    * - 2.x
       - Maintenance
       - End of Support
-      - 2022-07-08
-    * - 2.x
-      - General Availability
-      - 
-      - 
+      - 2023-03-02
     * - 3.x
-      - General Availability 
+      - General Availability
       - 
       - 
 
--- a/SUPPORT_POLICY.rst
+++ b/SUPPORT_POLICY.rst
@@ -22,16 +22,12 @@ This table describes the current support status of each major version of the AWS
       - Next status
       - Next status date
     * - 1.x
-      - Maintenance
       - End of Support
-      - 2022-07-08
+      - 
+      - 
     * - 2.x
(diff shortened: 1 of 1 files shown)


SUPPORT_POLICY.rst
```diff
@@ -22,16 +22,12 @@ This table describes the current support status of each major version of the AWS
...
+      - 
+      - 
     * - 2.x
       - General Availability
-      - Maintenance
-      - 2021-07-13
-    * - 3.x
-      -
-      - General Availability 
-      - 2021-07-13
```
josecorella, line 35: umm why are we removing this? we have a major version 3.x for the js esdk...

josecorella (changes requested):

SUPPORT_POLICY.rst
texastony: We do?

texastony: Wait... are you going to tell me that `mainline` is the branch for both v2 and v3?

josecorella: yup! https://github.com/aws/aws-encryption-sdk-javascript/blob/master/CHANGELOG.md#311-2022-03-15

texastony: oi...

texastony: In which case, both the 2.x and 3.x are in GA.

texastony: Resolved by a468061

SUPPORT_POLICY.rst
```diff
@@ -27,11 +27,11 @@ This table describes the current support status of each major version of the AWS
       - 2022-07-08
     * - 2.x
       - General Availability
-      - Maintenance
-      - 2021-07-13
```
josecorella, line 31: technically 2.x is in maintenance mode, when we released 3.x i never created the mainline-2.x branch...
we haven't had to do a fix for code on the 2.x version
we should probably create a 2.x branch

texastony: Oh... fudge... it never ends...
Are all the node modules individually versioned to 3.x?

josecorella: yes, that's not a problem though
we can just check out at this commit https://github.com/aws/aws-encryption-sdk-javascript/tree/05117803a1cc6e0d4fa67845eefabde6d2930980
and push a branch to upstream titled `mainline-2.x`

texastony: Alright. 
The "breaking change" for 3.x was removal of support for NodeJS 8 and 10.
But, too my knowledge, we are not running tests against 2.x at all...
So... is 2.x in maintenance, or is it in End-of-Support?

Because we have no mechanism validating the package-lock of 2.x is in a good state...

josecorella: you are correct, no tests are run against 2.x, it should be in maintenance

texastony: Resolved by e45a109

Issue #1034, opened 2022-10-17
Number increment for @aws-crypto/client-node


solotoo:
I have seen https://github.com/aws/aws-encryption-sdk-javascript/issues/1031 but it doesnt make any sense.
Incrementing from 3.1.1 to the latest being 1.9.1 is not how versions should increment.
The package page at https://www.npmjs.com/package/@aws-crypto/client-node give no explanation as to why the version number has gone down. Its confusing and a worry that we might have been seeing another fakerjs incident

texastony:
@solotoo 

I now see what @solotoo  and @lahuman are referring to.

We (AWS Crypto Tools) recently published a new minor version of the ESDK-JS-1.x.

NPM has registered that new version as the latest version.

That is not the case.

3.1.1 is the latest version; the patch we put out was only for the 1.x version.

Thank you @solotoo and @lahuman for bringing this to our attention.

We will address this gap in our release process and update the NPM page to mark 3.1.1 as `latest`.

justplaz:
We have updated npm to mark the correct version number as `latest` for this package and all other @aws-crypto packages which were incorrectly set to the most recently released `1.x` versions. 

Please do not hesitate to reach out if you encounter any future issues. Thanks!

PR #1037, opened 2022-10-24
fix: set dist-tag to the major version to avoid overwriting latest ta…

justplaz:
…g on non-latest MV branch

*Issue #, if available:* https://github.com/aws/aws-encryption-sdk-javascript/issues/1034 (long term fix)

*Description of changes:* See the above issue. When we ran a release to retire 1.x, no tag was specified, so the default 
 was set to `latest` ([doc ref](https://www.npmjs.com/package/@lerna/publish#--dist-tag-tag)), overwriting from the actual latest `3.x` release. The `1.x` release is now retired, but `2.x` is not yet retired and will eventually need to be released, upon retirement or otherwise. 

This change sets the `dist-tag` to `2.x` for the CodeBuild spec in `2.x` to avoid the situation which occurred during `1.x` retirement. 

# Check any applicable:

### DIFF
--- a/codebuild/release/prod-release.yml
+++ b/codebuild/release/prod-release.yml
@@ -57,7 +57,9 @@ phases:
       # This is going to use the OTP generated above and the NPM_TOKEN
       # environment variable. This will only publish things that are
       # missing from npm. It is therefore safe to run repeatedly.
-      - npx lerna publish from-package --yes --otp $OTP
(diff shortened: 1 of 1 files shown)


justplaz:
closing in favor of: https://github.com/aws/aws-encryption-sdk-javascript/pull/1038 (messed up my branches, apologies)

PR #1041, opened 2022-10-28
feat: Support Node v18

texastony:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

# Check any applicable:
- [ ] Were any files moved? Moving files changes their URL, which breaks all hyperlinks to the files.

PR #1043, opened 2022-10-30
feat: Support AWS SDK v3

seebees:
To support AWS SDK v3,
there needs to be a single interface
between AWS SDK v2 and v3
for the AWS KMS Client.
The `AwsEsdkKMSInterface` is a close copy
that needs to be shared between `kms-keyring`,
`kms-keyring-node`, and `kms-keyring-browser`.

Adding `@aws-sdk/client-kms` as a devDependencies
tests have been added for v2 and v3
for every AWS KMS keyring.

Two additional changes were required.

First, getting the `region` value
from an AWS SDK v3 client is an `async` function,
rather than a string property.
This means that the `AwsKmsMrkAwareSymmetricDiscoveryKeyring`
needs to be able to handle this.
The constructor, a synchronous function
needs to store enough state
that the _OnDecrypt function
can await the setting of the region value.

Second in setting this value `readOnlyProperty`
was updated to explicitly set `writable: false`.
This allows this helper to update a property
and set this updated value to read only.

# Check any applicable:


seebees:
This is built on #1041 and should be rebase after that has been merged

modules/kms-keyring/test/kms_mrk_discovery_keyring.constructor.test.ts
```diff
@@ -123,9 +123,10 @@ describe('AwsKmsMrkAwareSymmetricDiscoveryKeyring: constructor', () => {
     })
     expect(regionCalled).to.equal(true)
     expect(test.clientRegion).to.not.equal(client.config.region)
-    // @ts-ignore This is to check that the value is a promise
+    // This is to check that the value is a promise
     expect(
       typeof test.clientRegion === 'object' &&
+        // @ts-ignore
```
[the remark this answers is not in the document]
seebees: It should, since it checks the formatting every time.
This is also in the test.

brianfrantz:
Is there an ETA on pushing this out?  Guessing this will unblock me from using AWS SSO, which isn't working in the current version...

modules/kms-keyring-browser/test/kms_keyring_browser.test.ts
```diff
@@ -5,7 +5,9 @@
 
 import * as chai from 'chai'
 import chaiAsPromised from 'chai-as-promised'
-import { KmsKeyringBrowser, getClient, KMS } from '../src/index'
+import { KmsKeyringBrowser, getClient } from '../src/index'
+import { KMS as V2KMS } from 'aws-sdk'
+import { KMS as V3KMS } from '@aws-sdk/client-kms'
```
texastony, line 10: Question: 
If we do not need `import { KMS } from 'aws-sdk'` in any of the source files,
[should we remove `"aws-sdk"` from `kms-keyring-browser`'s `package.json`](https://github.com/aws/aws-encryption-sdk-javascript/blob/fd8264a3c4a0a34f8e554d56caedbdde2d7bf7d8/modules/kms-keyring-browser/package.json#L25)?

Or can we declare it a test-only dependency?

texastony (commented):
I am not Java Script savy enough to know if we can mark a dependency as a test or dev dependency inside a module.

If we cannot, does this mean that our customers will pull in both AWS SDKs?

modules/kms-keyring-browser/test/kms_keyring_browser.test.ts
seebees: Unfortunately we need to keep it for now.
This is because a customer may be getting their `aws-sdk` dependency from this package.

We can remove it on the next major version.

texastony (approved):

texastony:
Our Release Build failed on `testBrowser`:
```
[Container] 2023/02/23 20:29:03 Running command n 18
  installing : node-v18.14.2
       mkdir : /usr/local/n/versions/node/18.14.2
       fetch : https://nodejs.org/dist/v18.14.2/node-v18.14.2-linux-x64.tar.xz
     copying : node/18.14.2
node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by node)
/usr/local/bin/node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by /usr/local/bin/node)
   installed :  (with npm )

[Container] 2023/02/23 20:29:18 Running command npm ci --unsafe-perm
node: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by node)

[Container] 2023/02/23 20:29:18 Command did not exit successfully npm ci --unsafe-perm exit status 1
[Container] 2023/02/23 20:29:18 Phase complete: INSTALL State: FAILED
[Container] 2023/02/23 20:29:18 Phase context status code: COMMAND_EXECUTION_ERROR Message: Error while executing command: npm ci --unsafe-perm. Reason: exit status 1
```

Fudge... Maybe the CodeBuild Image is too old...

texastony:
> Is there an ETA on pushing this out? Guessing this will unblock me from using AWS SSO, which isn't working in the current version...

@brianfrantz 
We have released AWS SKD V3 support in ESDK-JS 3.2.0.
Note: The ESDK-JS still lists the AWS SDK V2 as a dependency.
We will release a new major version that removes this V2 dependency
as a breaking change.

luk2302:
@texastony any update on when that major release is supposed to happen?

texastony:
@luc2302, this work is on our roadmap.
However, we cannot commit to a date.

dmyerspgr:
Bumping - Hopeful AWS SDK v3 major release soon! This would greatly improve our lambdas performance and cost.

PR #1045, opened 2022-10-31
chore: Add dist-tag to release

seebees:
Currently the release build will _always_ add the `latest` tag to every release.
If a release is for an old version,
this will result in the wrong version being tagged as `latest`.

This ENV in the build allows for proper tagging.

# Check any applicable:

### DIFF
--- a/codebuild/release/publish.yml
+++ b/codebuild/release/publish.yml
@@ -4,6 +4,8 @@ env:
   variables:
     NODE_OPTIONS: "--max-old-space-size=4096"
     BRANCH: "master"
+    # An explicit distribution tag
+    DIST_TAG: "latest"
   secrets-manager:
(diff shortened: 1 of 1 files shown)


codebuild/release/publish.yml
```diff
@@ -4,6 +4,8 @@ env:
   variables:
     NODE_OPTIONS: "--max-old-space-size=4096"
     BRANCH: "master"
+    # An explicit distribution tag
+    DIST_TAG: "latest"
```
texastony, line 8: Question: Can this be overridden at some point before the release?
I think that is how CodeBuild works... but I just want to verify.

josecorella (approved):
lgtm - when we port this over to the 2.x branch is the value of `DIST_TAG` == 2.x?

seebees:
I think so.
This is _how_ to set the value,
not what the value should be for any given release.

PR #1178, opened 2023-06-29
feat!: Test & Support only latest NodeJS

texastony:
*Issue #, if available:* Drop Node 14

*Description of changes:* 
In CodeBuild Builds:
1. Always specify the exact Node version
2. Name files after the Node Version
3. Stop testing NodeJS 12 & 14
4. Start testing NodeJS 20

BREAKING CHANGE:
This library no longer tests against nor supports NodeJS 12 or 14.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

# Check any applicable:
- [ ] Were any files moved? Moving files changes their URL, which breaks all hyperlinks to the files.

PR #1177, opened 2023-06-29
feat!: Drop Support & Dependency on AWS SDK V2

texastony:
*Issue #, if available:* #1100, #922, #865

*Description of changes:* `Drop Support & Dependency on AWS SDK V2`

BREAKING CHANGE:
The AWS Encryption SDK for JavaScript:
- does not supports the AWS SDK for JavaScript V2
- requires the AWS SDK for JavaScript V3's kms-client (if using the KMS Keyring).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

# Check any applicable:
- [ ] Were any files moved? Moving files changes their URL, which breaks all hyperlinks to the files.

Issue #1100, opened 2023-02-28
KmsKeyringNode dependency on aws-sdk resulting in Deprecation WARNings


RalphBragg:
### Problem:

This library is still dependent on the full aws-sdk despite numerous tickets and requests being raised for its removal by this community (mainly for lambda use cases). I believe it is now the cause of the WARN messages being raised in any package and service that is using the node crypto library. Please advise on what the intended solution is for this please and the timelines. Happy to migrate to another capability if there's a migration guide available

(node:92870) NOTE: We are formalizing our plans to enter AWS SDK for JavaScript (v2) into maintenance mode in 2023.
Please migrate your code to use AWS SDK for JavaScript (v3).
For more information, check the migration guide at https://a.co/7PzMCcy
    at emitWarning (/Users/REDACTED/node_modules/aws-sdk/lib/maintenance_mode_message.js:21:13)
    at Timeout._onTimeout (/Users/REDACTED/node_modules/aws-sdk/lib/maintenance_mode_message.js:29:5)

### Solution:

Use something other than the full aws-sdk library

### Out of scope:

Is there anything the solution will intentionally NOT address?

[//]: #  (NOTE: If you believe this might be a security issue, please email aws-security@amazon.com instead of creating a GitHub issue. For more details, see the AWS Vulnerability Reporting Guide: https://aws.amazon.com/security/vulnerability-reporting/ )

JohnBrown0126:
Can we get some eyes on this

defy93:
+1 this issue as well

texastony:
@RalphBragg (+ @JohnBrown0126 , @defy93 , and others),

We are aware that the AWS SDK for JavaScript V2 is going into maintenance.
Recently, we released a new minor version of the AWS Encryption SDK for JavaScript (ESDK-JS),
[3.2.0, which featured support for the AWS SDK for JavaScript V3](https://github.com/aws/aws-encryption-sdk-javascript/blob/master/CHANGELOG.md#320-2023-02-23). 

We did not remove the AWS SDK V2 dependency, as we did not want break any current customers,
who may have relied on the V2 SDK existing in the ESDK-JS.

We will release the ESDK-JS that only relies on the AWS SDK V3 in a new major version.

(Leaving this issue open, pending the new Major Version release described above)

Much Obliged,
[AWS Crypto Tools](https://github.com/aws/crypto-tools)

revmischa:
[...]

Issue #922, opened 2022-05-24
@aws-crypto/kms-keyring-node dependency issue


d3mi3n:
### Problem:

`@aws-crypto/kms-keyring-node` imports `@aws-crypto/material-management` in [this file](https://github.com/aws/aws-encryption-sdk-javascript/blob/master/modules/kms-keyring-node/src/kms_mrk_strict_multi_keyring_node.ts#L9), but `@aws-crypto/material-management` is not listed as a dependency

### Solution:

Add `@aws-crypto/material-management` as a dependency of `@aws-crypto/kms-keyring-node` or (if possible) use `@aws-crypto/material-management-node` which is a dependency already.

ajewellamz:
https://github.com/aws/aws-encryption-sdk-javascript/pull/1018

trevorpoppen:
Hello, I'm seeing this dependency issue rear it's head again in `"@aws-crypto/client-node": "^3.2.0"` ;

Import throwing the error is  `import { NodeAlgorithmSuite } from '@aws-crypto/material-management';` from `@aws-crypto/kms-keyring-node`;

I'm using PlugnPlay in Yarn 3.3.0 and typescript 4.9.3

josecorella:
Hello @d3mi3n,

Major Version 4.0.0 of the AWS ESDK for JS has been released. This major version should address the problems you outline on this issue. Please let us know if you have any questions.

AWS Crypto Tools

PR #1180, opened 2023-06-30
feat!: Remove AWS SDK V2 Dependency

texastony:
*Issue #, if available:*  #1100, #922, #865, #1178, #1177

*Description of changes:*
In CodeBuild Builds:
1. Always specify the exact Node version
2. Name files after the Node Version
3. Stop testing NodeJS 12 & 14
4. Start testing NodeJS 20
5. Use Node 18 for compliance

BREAKING CHANGE:
The AWS Encryption SDK for JavaScript:
- requires the AWS SDK for JavaScript V3's kms-client (if using the KMS Keyring).
- no longer requires the AWS SDK V2
- no longer tests against nor supports NodeJS 12 or 14.

Note: @seebees I had to merge #1177 and #1178 to get CodeBuild to pass.
I think the new package-lock only supports Node >= 16, 
so I needed to drop the Node 14 tests. 

# Check any applicable:


texastony:
I do not know why, but the CB `testVectorsNodejs20` completes `npm run verdaccio-publish` and then just hangs until the CB request times out...

ajewellamz (dismissed):

util/local_verdaccio_publish
```diff
@@ -35,21 +41,33 @@ const args = [
...
+    console.log(`killing Verdaccio returned ${verdaccioKilledStatus}`);
+    
     // If this command had an error,
     // we need to forward this.
     // Otherwise the entire CI build may think that things succeeded.
     if (code !== 0) throw Error(`Exit code: ${code}`)
-  })
\ No newline at end of file
+
+    process.exit()
```
seebees, line 71: My guess is that this is required because you have added a `timeout`.
That timeout is likely implemented with `setTimeout`.
This means when this is executed there is an event in the background.

seebees (dismissed):
LGTM

util/local_verdaccio_publish
texastony: Hmm... it was hanging before I added the timeout. 
But I guess I could check to see if just exit fixes the hang, with out the time out.

lucasmcdonald3 (approved):

PR #1222, opened 2023-08-30
feat: update to remove security vulnerabilities

ajewellamz:
*Issue #, if available:*

*Description of changes:*

# Check any applicable:


package-lock.json
texastony, line 1: What commands did you run to update the package lock?

I ask, as package.json was not updated, only the lock.
Updates only to the package lock, not the package.json,
can happen and be valid.

But I want to be sure.

PR #1580, opened 2025-02-26
feat: integration-node can produce decrypt manifests

seebees:
From a given encrypt manifest,
integration-node now takes
`—-decryptManifest` | `-d`

This will create a ziped decrypt manifest
that can be consumed by decrypt.

# Check any applicable:

### DIFF
--- a/modules/integration-node/src/integration_tests.ts
+++ b/modules/integration-node/src/integration_tests.ts
@@ -183,9 +183,9 @@ function composeEncryptResults(
       },
       manifestZip: manifest.manifestZip,
     }
-  } else if (!!decryptOracle) {
+  } else if (decryptOracle) {
     return decryptOracleEncryptResults(decryptOracle)
-  } else if (!!decryptManifest) {
+  } else if (decryptManifest) {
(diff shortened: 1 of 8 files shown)


modules/integration-node/src/integration_tests.ts
```diff
@@ -134,10 +218,69 @@ export async function testEncryptVector(
...
+}
+
+function decryptionManifestEncryptResults(
+  manifestPath: string
+): ProcessEncryptResults {
+  const manifestZip = new ZipFile()
+  const manifest: DecryptManifestList = {
+    manifest: {
+      type: 'awses-decrypt',
+      version: 1,
```
seebees, line 235: Change this to version 2 plz

modules/integration-node/src/integration_tests.ts
```diff
@@ -134,10 +218,69 @@ export async function testEncryptVector(
...
+    },
+    client: {
+      name: 'aws/aws-encryption-sdk-javascript',
+      //Get the right version
+      version: '2.2.0',
+    },
+    keys: 'file://keys.json',
```
seebees, line 242: Add comment

PR #1638, opened 2026-03-05
fix(ci-auth): git release auth

ShubhamChaturvedi7:
*Issue #, if available:*

*Description of changes:*

# Check any applicable:

### DIFF
--- a/.github/workflows/prod-release.yml
+++ b/.github/workflows/prod-release.yml
@@ -79,6 +79,20 @@ jobs:
           git config --global user.email "no-reply@noemail.local"
           git checkout $BRANCH
 
+      - name: Verify git auth identity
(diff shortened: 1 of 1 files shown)


.github/workflows/prod-release.yml
```diff
@@ -37,6 +37,7 @@ jobs:
         with:
           fetch-depth: 0
           submodules: true
+          persist-credentials: false
```
josecorella: dont we also need to use the github secret now?

josecorella (approved):

.github/workflows/prod-release.yml
Adam9911: adam9911

PR #1662, opened 2026-06-08
fix: Using discovery filter while constructing keyring

sharmabikram:
*Issue #, if available:*

*Description of changes:*
The discovery filter was not getting propagated and was dropped. This change ensures that filters are not getting dropped.

# Check any applicable:

### DIFF
--- a/modules/kms-keyring-browser/test/kms_keyring_browser.test.ts
+++ b/modules/kms-keyring-browser/test/kms_keyring_browser.test.ts
@@ -55,6 +55,95 @@ describe('KmsKeyringBrowser::constructor', () => {
     expect(test.isDiscovery).to.equal(true)
     expect(test.discoveryFilter).to.deep.equal(discoveryFilter)
   })
+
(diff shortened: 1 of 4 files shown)


modules/kms-keyring-browser/test/kms_keyring_browser.test.ts
```diff
@@ -48,6 +48,13 @@ describe('KmsKeyringBrowser::constructor', () => {
     const test = new KmsKeyringBrowser({ discovery: true })
     expect(test instanceof KeyringWebCrypto).to.equal(true)
   })
+
+  it('forwards discoveryFilter to base KmsKeyring', () => {
```
kessplas, line 52: we should add the other test from the node keyring here as well

also, we should test that an invalid filter throws on create when discovery is set to true

kessplas (approved):

modules/kms-keyring-browser/test/kms_keyring_browser.test.ts
sharmabikram: Added more tests

PR #1680, opened 2026-07-22
fix(release): publish on Node LTS so npm@latest installs

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*

Bump the publish job's Node to always publish with `lts/*`.

The latest npm (12.x) requires Node `^22` so this would work, but it probably makes sense to always use LTS to publish (also keeping in line with this PR to the 4.x branch https://github.com/aws/aws-encryption-sdk-javascript/pull/1679)

# Check any applicable:

### DIFF
--- a/.github/workflows/prod-release.yml
+++ b/.github/workflows/prod-release.yml
@@ -103,7 +103,7 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: '22.22.1'
+          node-version: 'latest'
           registry-url: 'https://registry.npmjs.org'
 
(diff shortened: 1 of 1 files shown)


.github/workflows/prod-release.yml
```diff
@@ -103,7 +103,7 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: '22.22.1'
+          node-version: 'latest'
```
lucasmcdonald3:
```suggestion
          node-version: 'lts/*'
```

kessplas (approved):

